PR #1, opened 2017-06-11
Here's a building Vts.NetStandard library project.

dcuccia:
- Needs VS 2017 Preview 1 or later (see instructions here: https://www.microsoft.com/net/core#windowsvs2017)
- Note that Silverlight is no longer supported in VS 2017, at least for now (boo, hiss)
- Haven't dealt with specifying build actions/types for resource files
- Updated the project to use the latest MathNet.Numerics (3.19.0)
- Had to update two lines of code: 1) unused reference to Monte Carlo lib from IForwardSolver.cs, and 2) update Linear interp call in ReffCalculator.cs
- Could be drop-in replacement for the new WPF library...worth a try, once the references are embedded correctly
- Haven't figured out what happens when the NuGet dependencies used don't support the netstandard2.0 profile for non-desktop (.NET 4.7) runtimes, such as .NET Core on Windows/Mac/Linux/Xamarin.
- Haven't used the API Port tool yet, which will shed light on the bullet above: https://github.com/Microsoft/dotnet-apiport

### DIFF
--- a/src/Vts.MonteCarlo.CommandLineApplication.Test/app.config
+++ b/src/Vts.MonteCarlo.CommandLineApplication.Test/app.config
@@ -1,23 +1,23 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <runtime>
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
       <dependentAssembly>
-        <assemblyIdentity name="Microsoft.Practices.ServiceLocation" publicKeyToken="31bf3856ad364e35" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-1.2.0.0" newVersion="1.2.0.0" />
+        <assemblyIdentity name="Microsoft.Practices.ServiceLocation" publicKeyToken="31bf3856ad364e35" culture="neutral"/>
+        <bindingRedirect oldVersion="0.0.0.0-1.2.0.0" newVersion="1.2.0.0"/>
       </dependentAssembly>
       <dependentAssembly>
-        <assemblyIdentity name="System.Reactive.Core" publicKeyToken="31bf3856ad364e35" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-2.2.4.0" newVersion="2.2.4.0" />
+        <assemblyIdentity name="System.Reactive.Core" publicKeyToken="31bf3856ad364e35" culture="neutral"/>
+        <bindingRedirect oldVersion="0.0.0.0-2.2.4.0" newVersion="2.2.4.0"/>
       </dependentAssembly>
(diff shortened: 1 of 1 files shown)


dcuccia:
Just a note - I pulled this on a new computer, and wasn't able to build the netstandard2.0 library. Error: "One or more packages are incompatible with .NETStandard,Version=v2.0."		

This honestly is more what I was assuming would happen. Per the answer here, it has to be netstandard "all the way down":

https://stackoverflow.com/questions/37710474/one-or-more-packages-are-incompatible-with-netstandard-version-v1-5

The only obvious solution I see is to see if there are netstandard2.0 implementations of the NuGet dependencies in their alpha/beta channels, or to build them ourselves until they're available for netstandard2.0.

dcuccia:
Looks like it my errors may have something to do with having VS 2017 15.2 installed, instead of VS 2017 15.3 Preview. Here's the link to 15.3: https://www.visualstudio.com/vs/preview/

dcuccia:
...nope, same problems as before.

lmalenfant:
We have converted the Silverlight Vts library and Test project to .NET Standard 2.0 and .NET Core and archived the Silverlight version due to it not being supported in Visual Studio 2017. This allowed us to keep all the source files in place with the Vts.Desktop project retaining the links. The .NET Standard version builds and all the unit tests pass. We have a few issues with the built in Visual Studio test runner but Resharper works great, as does "dotnet test" from the command line.
There is one library DotNetZip that doesn't have a .NET Standard version so it pulls in the 4.6.1 .NET framework version.

dcuccia:
Very nice! A few questions:

1) Did you need to use the Windows Compatibility Pack to get this to work,
or is it "vanilla" .NET Standard 2.0? (In other words, will it run on Linux
and Mac too?)

2) Does DotNetZip run on the .NET Core runtime, even though the library was
originally built for the 4.6.1 runtime? (If not, I see here
<https://github.com/icsharpcode/SharpZipLib/issues/106> that SharpZipLib
supports .NET Standard 1.3 and above).

______________________________________
David Cuccia
CEO & CTO
david.cuccia@modulatedimaging.com
949.596.7492
Skype: david.cuccia.mi

lmalenfant:
It is "Vanilla" .NET Standard 2.0 but we haven't tried to run it on Linux or Mac other than with Mono. This will be the next step.

I haven't investigated DotNetZip but we actually don't use it anymore because it was used in the Silverlight application for saving files. The WPF application no longer zips the files. Thank you for the link to SharpZipLib, that might be worth checking out since it is .NET Standard already.

dcuccia:
That's great news. On Linux/Mac, you'll have the choice of Mono, .NET Core,
or even Xamarin as runtimes, as they now both implement .NET Standard.

lmalenfant:
We have converted our Silverlight library to .NET Standard 2.0, closing this pull request.

PR #45, opened 2022-08-12
Feature/lm spell check

lmalenfant:


hayakawa16 (approved):
I found one edit that should be reversed
ForLisa
Otherwise I approve all other modifications.

lmalenfant:
I reverted this back by adding in the references for NaN and Infinity so it will look like this in the documentation:
image
The types NaN, PositiveInfinity and NegativeInfinity are clickable links to the Microsoft documentation.

PR #53, opened 2022-11-07
Modifications to fix bug that counted number of photons killed by Rus…

hayakawa16:
…sian Roulette incorrectly.  There is no time schedule for this fix now so no need to review quickly.

### DIFF
--- a/src/Vts/MonteCarlo/Extensions/StatisticsExtensions.cs
+++ b/src/Vts/MonteCarlo/Extensions/StatisticsExtensions.cs
@@ -8,41 +8,46 @@ namespace Vts.MonteCarlo.Extensions
     public static class StatisticsExtensions
     {
         /// <summary>
-        /// Method to determine statistics about how the photon died.
+        /// Method to determine statistics about how the photon died.  These states
+        /// are typically mutually exclusive, however there are occurrences of photons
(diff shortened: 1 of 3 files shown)


src/Vts/MonteCarlo/Extensions/StatisticsExtensions.cs
```diff
@@ -8,41 +8,46 @@ namespace Vts.MonteCarlo.Extensions
     public static class StatisticsExtensions
     {
         /// <summary>
-        /// Method to determine statistics about how the photon died.
+        /// Method to determine statistics about how the photon died.  These states
+        /// are typically mutually exclusive, however there are occurrences of photons
+        /// that are killed by RR at start of track that crosses boundary
+        /// and gets that state set as well but does not tally.
+        /// It is for this reason the check for RR is first.
         /// </summary>
```
lmalenfant, line 16: Do you think we should write out Russian roulette in this documentation?

lmalenfant (approved):

src/Vts/MonteCarlo/Extensions/StatisticsExtensions.cs
hayakawa16: Sounds good.  Thanks for suggestion.

PR #56, opened 2022-12-09
Lm add unit tests and cleanup

lmalenfant:
The unit tests are up to 100% and the code cleanup is done. I pulled in master and the .NET 4.8 unit tests run fine.

### DIFF
--- a/src/Vts.Test/Factories/ComputationFactoryTests.cs
+++ b/src/Vts.Test/Factories/ComputationFactoryTests.cs
@@ -661,6 +661,90 @@ namespace Vts.Test.Factories
             Assert.IsTrue(Math.Abs(fluence[0].Real - 0.188294) < 0.000001);
         }
 
+        [Test]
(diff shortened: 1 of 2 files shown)


hayakawa16 (approved):
What a extensive effort!  Had questions about the use of "IReadOnlyList<Complex>" (ComputationFactory line 87) compared to prior "Complex[]": have not come across IReadOnlyList before, why is it warranted and why change from array to IList?

lmalenfant:
> What a extensive effort! Had questions about the use of "IReadOnlyList" (ComputationFactory line 87) compared to prior "Complex[]": have not come across IReadOnlyList before, why is it warranted and why change from array to IList?

It was a clean-up suggestion, probably one of the ones that prompted me to increase code coverage before making changes because I wanted to make sure it didn't break anything. I have used it before, it is my understanding that it is good to use "readonly" where possible because it makes the code more readable, another developer would know that the value will not change after it is set.

PR #60, opened 2023-01-11
Feature/update documentation

lmalenfant:


hayakawa16 (approved):
I think I know why some of the "Deleted" and "Added" plots of the vts_mc_demo are not identical (for example vts_mc_demo_04.png).  I checked and the script calls the default constructor for SimulationInput which calls SimulationOptions with Seed=-1 (new seed each time run).  I guess that is okay, just caught my eye.

lmalenfant:
> I think I know why some of the "Deleted" and "Added" plots of the vts_mc_demo are not identical (for example vts_mc_demo_04.png). I checked and the script calls the default constructor for SimulationInput which calls SimulationOptions with Seed=-1 (new seed each time run). I guess that is okay, just caught my eye.

Yes, I think the plots are numbered based on the order they appear so if we add one it might appear that it replaces an unrelated plot. As long as it looks correct in the documentation, we should be ok. I noticed that too in the Files Changed tab.

hayakawa16:
I was thinking that perhaps I should seed the simulations so that the plots look the same each time we run them regardless of the order.  I will put this on my "todo" list.

PR #62, opened 2023-03-31
Fixed the bug and the vunerabilities identified by Sonar Cloud

lmalenfant:


hayakawa16 (approved):
The changes to the Monte Carlo code make sense to me.  They are in very basic methods, so if the unit tests are passing, they should be good.

lmalenfant:
I updated the NuGet packages and pulled in the WPF to run the unit tests. Everything worked fine.

PR #66, opened 2023-05-31
Ch 230221 more mods for fluorescence

hayakawa16:


hayakawa16:
Hi, I'm ready for the review of this branch.  I added Janaka.  
Janaka, there is an easy way to review changes by using "Files changed" tab above which shows changes from Master.  Let me know if you have questions.  Thanks for your help!

janakarana:
"There are two methods used to sample the absorbed energy: CDF or Uniform.  Using the CDF method, a probability density function (PDF) and resulting cumulative  function (CDF) is determined from the absorbed energy map.  Then the number of photons  launched from each fluorescent voxel is determined probabilistically from sampling the CDF.      And the weight of each photon is set to the total absorbed energy of the fluorescing region because if N photons are launched the results are normalized by N producing back the total absorbed energy. Using the Uniform method, the same number of photons is sent out from each voxel with weight equal to the absorbed energy of that voxel.  HOWEVER, the results of a multiple voxel source CANNOT be combined because the initial weight is different  So if the fluorescing region has more than 1 voxel, each voxel must be simulated with a different simulation (possibly using the parallel MC)."

I  fully agree on this approach. I think you have to normalize the results by the volume of the source voxel. Because if you consider a non-cuboidal shape, only part of the  boundary voxels contribute.  If you include the volume of the voxel during absorbed energy calculation, it can be ignored.

janakarana (approved):

hayakawa16:
Thank you very much for your review Janaka!  If I understand your last comment correctly, then yes the absorbed energy results read in by the fluorescent source have already been normalized by the volume in which the tally was performed.  So for AOfXAndYAndZ the results were written to the file normalized by the voxel volume and for AOfRhoAndZ the results were written to file normalized by the respective cylindrical volume.

One of the things I learned the hard way, is that our absorbed energy tallies tally photon contributions beyond the grid specified by the detector into the last bin.  We did this so that multiplying back by the bin size and summing would produce total absorbed energy.  These "edge" bins cannot be included in the absorbed excitation because of this.  So specifying the appropriate domain for the absorbed energy tallies (adding a extra bin on the edges) needs to be done because in the fluorescent source code, I omit these edge voxels.

hayakawa16:
Lisa reviewed the validation code and saw the use of the "dynamic" cast can be replaced with actual class cast and should.  I plan to do this and have a re-review from Lisa.

hayakawa16:
In the validation classes, I replaced the "dynamic" case to actual class when possible.  In addition, I made all validation classes "static".  Ready for re-review.

src/Vts/MonteCarlo/DataStructuresValidation/SimulationInputValidation.cs
```diff
@@ -61,10 +61,16 @@ private static ValidationResult ValidateSourceInput(ISourceInput sourceInput, IT
                     "Source input not valid given tissue definition",
                     "Alter sourceInput.InitialTissueRegionIndex to be consistent with tissue definition");
             }
+            if (sourceInput is FluorescenceEmissionAOfRhoAndZSourceInput)
+                return FluorescenceEmissionAOfRhoAndZSourceInputValidation.ValidateInput(sourceInput);
```
lmalenfant, line 65: Did we decide that if we inline the return then we wouldn't need curly braces, this is not exactly inline, it's on the next line which could lead to bugs if somebody forgets the braces. I know these lines are long but should we inline these?

src/Vts/MonteCarlo/DataStructuresValidation/SimulationInputValidation.cs
```diff
@@ -79,31 +85,22 @@ private static ValidationResult ValidateTissueInput(ITissueInput tissueInput)
                      "Tissue optical properties mua, mus', n need to be non-negative",
                      "Please check optical properties");
             }
+
             if (tissueInput is MultiLayerTissueInput)
-            {
                 return MultiLayerTissueInputValidation.ValidateInput(tissueInput);
```
lmalenfant, line 90: Same comment as above for these, should we inline them?

lmalenfant: I brought this up in Visual Studio and it is suggesting switch statements for these, maybe we could leave these now but change them to switch expressions once we move fully to .NET 6?

src/Vts/MonteCarlo/DataStructuresValidation/SourceInputs/FluorescenceEmissionAOfRhoAndZSourceInputValidation.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// Also, warns that if Uniform sampling is specified, only one
+        /// fluorescing cylindrical voxel can be simulated
+        /// </summary>
+        /// <param name="input">source input in SimulationInput</param>
+        /// <returns>An instance of the ValidationResult class</returns>
+        public static ValidationResult ValidateInput(ISourceInput input)
+        {
+            if (((FluorescenceEmissionAOfRhoAndZSourceInput)input).SamplingMethod != SourcePositionSamplingType.Uniform)
+                return new ValidationResult(true, "");
+
```
lmalenfant, line 29: Inline or add braces

src/Vts/MonteCarlo/DataStructuresValidation/SourceInputs/FluorescenceEmissionAOfXAndYAndZSourceInputValidation.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// this simulation and checks for the validity of the excitation results.
+        /// Also, warns that if Uniform sampling is specified, only one
+        /// fluorescing voxel can be simulated
+        /// </summary>
+        /// <param name="input">source input in SimulationInput</param>
+        /// <returns>An instance of the ValidationResult class</returns>
+        public static ValidationResult ValidateInput(ISourceInput input)
+        {
+            if (((FluorescenceEmissionAOfXAndYAndZSourceInput)input).SamplingMethod != SourcePositionSamplingType.Uniform)
+                return new ValidationResult(true, "");
```
lmalenfant, line 28: Inline or add braces.

src/Vts/MonteCarlo/DataStructuresValidation/SimulationInputValidation.cs
hayakawa16: I decided to use curly braces instead of inline because to be consistent with code above it which could be inline but uses curly braces.

src/Vts/MonteCarlo/DataStructuresValidation/SimulationInputValidation.cs
hayakawa16: I used curly braces again.  
Yes!  Going to switch in .NET 6 makes sense.  Here since the expression uses "is" cannot use switch directly here yet.

src/Vts/MonteCarlo/DataStructuresValidation/SourceInputs/FluorescenceEmissionAOfRhoAndZSourceInputValidation.cs
hayakawa16: Used curly braces.

src/Vts/MonteCarlo/DataStructuresValidation/SourceInputs/FluorescenceEmissionAOfXAndYAndZSourceInputValidation.cs
hayakawa16: Used curly braces.

PR #67, opened 2023-06-02
Fixed the BuildTestRelease so it runs unit tests for MCCL and MCPP

lmalenfant:
Updated the configurations in the solution file


hayakawa16 (approved):
Looks good!

PR #71, opened 2023-07-27
Update Unit Tests

lmalenfant:
Updated the unit tests so embedded resources can be accessed even if the assembly name does not match the project name.

These are the changes that I made to the unit tests in the MCCL project so the embedded resources could be accessed.


hayakawa16 (approved):
Great work Lisa!

PR #77, opened 2023-07-29
Changed the targeting to .NET 6.0 closes #76

### LINKED ISSUE #76: Remove the .NET Standard support from the VTS library
We would like to remove the support for .NET Standard 2.0 as it is no longer required for the MATLAB interoperability or the WPF application. The WPF application is now on a Windows version of .NET 6.0.

This will improve the running of unit tests, currently because we target .NET Standard and .NET 6.0 we have unit tests targeting .NET Framework 4.8 and .NET 6.0 so there are twice as many unit tests then we need.

This change will also allow us to move forward with the C# version we are using and we will be able to modernize our code to C# version 10.

lmalenfant:
This was a very simple change but now the BuildTestRelease runs in about half the time.

@hayakawa16 I would recommend we get this merged next week before you pull master into your branches.

### DIFF
--- a/src/Vts.FemModeling/Vts.FemModeling.csproj
+++ b/src/Vts.FemModeling/Vts.FemModeling.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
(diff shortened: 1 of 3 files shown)


hayakawa16 (approved):
Looks great!

PR #85, opened 2023-08-14
Add resiliency to CustomBinaryWriterOfT.cs by creating any required directories first

### LINKED ISSUE #84: Add resiliency to CustomBinaryStreamWriterOfT
Currently, if the "results" folder is not written ahead of time, `PhotonDatabaseWriter` will throw an exception when attempting to write to a file within the specified folder. When consuming the Vts library, it is not obvious that this folder must exist before a simulation is run, and leads to exceptions that a consumer of the library should not have to debug/fix.

In order to fix this, add the following guard to the top of `CustomBinaryStreamWriter<T>.OpenStream()`:

```csharp
try
{
    // guard against directory not existing ahead of time
    var directory = Path.GetDirectoryName(_filename);
    if (!Directory.Exists(directory))
    {
        Directory.CreateDirectory(directory);
    }
    ...
}
```

dcuccia:
Fixes #84

### DIFF
--- a/src/Vts.Test/IO/CustomBinaryStreamWriterOfTTests.cs
+++ b/src/Vts.Test/IO/CustomBinaryStreamWriterOfTTests.cs
@@ -1,0 +1,50 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using NUnit.Framework;
+using Vts.IO;
+
+namespace Vts.Test.IO
+{
+    /// <summary>
+    /// Stream finder tests
+    /// </summary>
+    [TestFixture]
+    public class CustomBinaryStreamWriterOfTTests
+    {
+        /// <summary>
+        /// list of temporary files created by these unit tests
+        /// </summary>
+        readonly List<string> listOfTestGeneratedFolders = new List<string>()
+        {
+            "results_CustomBinaryStreamWriterOfTTests",
+        };
+
+        /// <summary>
+        /// clear all generated folders and files
+        /// </summary>
+        [OneTimeSetUp]
+        [OneTimeTearDown]
+        public void clear_folders_and_files()
+        {
+            foreach (var folder in listOfTestGeneratedFolders)
+            {
+                Directory.Delete(folder, recursive: true);
+            }
+        }
+
+        /// <summary>
+        /// Validate writing text to a stream
+        /// </summary>
+        [Test]
+        public void validate_does_not_throw_when_no_folder_exists()
+        {
+            // constructor opens the stream directly
+            Assert.DoesNotThrow(()=>
+            {
+                using var sut = new CustomBinaryStreamWriter<double>(
+                    fileName: @"results_CustomBinaryStreamWriterOfTTests/Nested/myfile.txt",
+                    writeMap: (bw, nextValue) => bw.Write(37));
+            });
+        }
+    }
+}
--- a/src/Vts.Test/IO/CustomBinaryStreamWriterOfTTests.cs
+++ b/src/Vts.Test/IO/CustomBinaryStreamWriterOfTTests.cs
@@ -28,6 +28,8 @@ namespace Vts.Test.IO
         {
             foreach (var folder in _listOfTestGeneratedFolders)
             {
+                if(!Directory.Exists(folder))
+                    continue;
                 Directory.Delete(folder, recursive: true);
             }
         }
--- a/src/Vts.Test/IO/CustomBinaryStreamWriterOfTTests.cs
+++ b/src/Vts.Test/IO/CustomBinaryStreamWriterOfTTests.cs
@@ -14,7 +14,7 @@ namespace Vts.Test.IO
         /// <summary>
         /// list of temporary files created by these unit tests
         /// </summary>
-        readonly List<string> listOfTestGeneratedFolders = new List<string>()
+        private readonly List<string> _listOfTestGeneratedFolders = new List<string>()
         {
             "results_CustomBinaryStreamWriterOfTTests",
         };
@@ -24,9 +24,9 @@ namespace Vts.Test.IO
         /// </summary>
         [OneTimeSetUp]
         [OneTimeTearDown]
-        public void clear_folders_and_files()
+        public void Clear_folders_and_files()
         {
-            foreach (var folder in listOfTestGeneratedFolders)
+            foreach (var folder in _listOfTestGeneratedFolders)
             {
                 Directory.Delete(folder, recursive: true);
             }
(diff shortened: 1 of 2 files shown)


src/Vts.Test/IO/CustomBinaryStreamWriterOfTTests.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <summary>
+    /// Stream finder tests
+    /// </summary>
+    [TestFixture]
+    public class CustomBinaryStreamWriterOfTTests
+    {
+        /// <summary>
+        /// list of temporary files created by these unit tests
+        /// </summary>
+        readonly List<string> listOfTestGeneratedFolders = new List<string>()
```
lmalenfant, line 17: Suggested name of a private field is _listOfTestGeneratedFolders and should be explicitly marked as private.

The type specification is not needed: new()

src/Vts.Test/IO/CustomBinaryStreamWriterOfTTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            "results_CustomBinaryStreamWriterOfTTests",
+        };
+
+        /// <summary>
+        /// clear all generated folders and files
+        /// </summary>
+        [OneTimeSetUp]
+        [OneTimeTearDown]
+        public void clear_folders_and_files()
```
lmalenfant, line 27: Method names should start with uppercase: Clear_folders_and_files

src/Vts.Test/IO/CustomBinaryStreamWriterOfTTests.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                Directory.Delete(folder, recursive: true);
+            }
+        }
+
+        /// <summary>
+        /// Validate writing text to a stream
+        /// </summary>
+        [Test]
+        public void validate_does_not_throw_when_no_folder_exists()
```
lmalenfant, line 39: Method name starts with uppercase: Validate_does_not_throw_when_no_folder_exists()

src/Vts/IO/CustomBinaryStreamWriterOfT.cs
```diff
@@ -72,6 +72,9 @@ private void OpenStream()
         {
             try
             {
+                // guard against directory not existing ahead of time
+                Directory.CreateDirectory(Path.GetDirectoryName(_filename));
```
lmalenfant, line 76: Possible 'null' assignment to non-nullable entity

dcuccia: Thanks - this one seems maybe a little out of scope - null assignment is rampant across the solution, and I'm just updating a method here where `_filename` is used in the same fashion. But I went ahead and enabled null type checking for the file and updated the code accordingly. Guess it doesn't hurt to update a whole file when we touch it.

src/Vts.Test/IO/CustomBinaryStreamWriterOfTTests.cs
dcuccia: Does it make sense to have a different standard for tests? Not sure there's a benefit of a test being capitalized for just the first word. Most of our tests in this project start with lowercase, at the moment.

lmalenfant: We have been updating them as we go along, we're just trying to follow a standard with all of the code going forward.

src/Vts/IO/CustomBinaryStreamWriterOfT.cs
lmalenfant: Since we started cleaning up the code, we are making a point to correct a file if we work on it. Due to the size of code base, it's hard to get to everything unless we are working on it. There are 2 more suggestions at the top of this file, to make _filename and _writeMap readonly.

lmalenfant:
I'm good with the changes, I would like to allow @hayakawa16 time to review and approve and then we are good to merge.

src/Vts/IO/CustomBinaryStreamWriterOfT.cs
dcuccia: I thought I had addressed those in my changeset but can double check in the morning

lmalenfant: The readonly is good now, it's just the IDisposable implementation that's left.

hayakawa16:
I have reviewed changes and all looks good to me.

hayakawa16 (approved):

lmalenfant (approved):

dcuccia:
I made those final changes for IDisposable, improved some null checks further, and fixed failing units tests. Ok to merge?

lmalenfant:
Nice @dcuccia! Going above and beyond, I just pulled the code and saw the failing unit test, let me do a pull and one last sanity check and then we can merge.

src/Vts/IO/CustomBinaryStreamWriterOfT.cs
```diff
@@ -10,11 +11,11 @@ namespace Vts.IO
...
-        private string _filename;
-        private Action<BinaryWriter, T> _writeMap;
-
-        private Stream _stream;
-        private BinaryWriter _binaryWriter;
+        private readonly string _filename;
+        private readonly Action<BinaryWriter, T> _writeMap;
+        private Stream? _stream;
+        private BinaryWriter? _binaryWriter;
+        private bool disposedValue;
```
lmalenfant, line 18: This name should be _disposedValue - if you merge before seeing this, we can fix later.

src/Vts/IO/CustomBinaryStreamWriterOfT.cs
```diff
@@ -114,91 +127,40 @@ public void Write(IEnumerable<T> items)
...
-        // If disposing equals false, the method has been called by the
-        // runtime from inside the finalizer and you should not reference
-        // other objects. Only unmanaged resources can be disposed.
-        private void Dispose(bool disposing)
+        /// <param name="disposing">Specifies whether the object is currently being disposed</param>
+        protected virtual void Dispose(bool disposing)
         {
-            // Check to see if Dispose has already been called.
-            if (!this._disposed)
+            if (!disposedValue)
```
lmalenfant, line 141: Invert "if" to reduce nesting - if you merge before seeing this, we can fix later.

dcuccia: Fixed :)

Issue #72, opened 2023-07-27
Implement and test script-style use of the VTS


dcuccia:
#### Main Idea
In order to lay the groundwork for future Jupyter Notebook and .NET Polyglot Notebook sample code, as envisioned in #70, this ticket is to create sample C# "scripts" ([top-level statements](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/top-level-statements)) that mimic the existing [VTS Matlab scripts](https://github.com/VirtualPhotonics/VTS/tree/master/matlab/vts_wrapper). In addition to unblocking the future work, at the end of this, anyone more comfortable working out of the VS or VS Code IDE should be able to use these scripts directly for their work if they desire.

#### Details
Design details:

1. Create these in a new project within the main `Vts` solution, with project name & namespace `Vts.Scripting`. This will ensure building the solution catches any build-level regressions to the scripts that might occur when the VTS library changes.
2. Use C# top-level statements combined with [SmallSharp](https://github.com/devlooped/SmallSharp) for simplifying running each script independently.
3. Make the code as "distraction free" as possible, removing as much unhelpful boilerplate code as possible, as these will be mirrored in other toolchains and languages.
4. Scripts will directly reference the `Vts.csproj` project within the `Vts` solution, to facilitate full use of the tools for design-time compilation error checking and debugging. This will differ from the future "external" Notebooks, which will resolve `Vts.dll` from Nuget.
5. Add a `Vts.Scripting.Test` project to cover the new script samples to catch/flag any runtime issues/bugs that will necessarily need updates to the libraries in the other repos

#### Acceptance Criteria
- Project should run "out of the box" with either VS or VS Code, requiring only a recently released version of .NET installed
- Should not require "classic" .NET Framework installed to run
- Should not concern the user with unnecessary details (e.g. throw logging errors, etc).
- Test project should successfully execute each script without errors

lmalenfant:
To follow naming convention of the other projects, the test application should be named Vts.Scripting.Test and it needs to be created now alongside this project. It is much easier to write unit tests at the same time as the code being tested.

As per our SonarCloud analysis, at least 80% code coverage is what we require for new code.

dcuccia:
Sounds good, thanks @lmalenfant !

dcuccia:
[...]

PR #86, opened 2023-08-16
72 Implement and test script-style use of the VTS

dcuccia:
Initial translated samples that demo "script-style" use the Vts, like in the Matlab scripts, as a foundation for consumption in .Net Polyglot and Jupyter Notebooks. For now, all 12 Monte Carlo demos are implemented and 11/18 of the Forward Solver demos are implemented.

Commented [here](https://github.com/VirtualPhotonics/VTS/issues/72#issuecomment-1679949865) to see how much more we ultimately want to do. IMO those could be added now, or at some time in the future. It is highly likely that these will need to evolve anyway _once_ we implement the notebooks, as we'll want to keep some level of parity on how the data is being sliced/diced and plotted. So, I suggest we consider having additional changes to these in the core library tracked in a separately Item, such as "Enhance and align script demos to those used in notebooks."

Note: It's unclear how/when we should include the Vts.Scripting.Tests in a CI process - it should probably be part of an integration test step, but we don't have that layer of granularity (yet - it's something I'd like to help with). Like the Matlab scripts they currently open up new windows that need to be closed. For now, just knowing that these libraries compile, and that the script tests pass when run ad-hoc is ok with me.


lmalenfant:
@dcuccia I notice that you are still pushing commits to this PR, I don't want to have to review it multiple times, could you let us know when it is finished and ready for review?

dcuccia:
Sorry - will do. When you mentioned that you'd probably take a few nights to get to it, I thought I'd squeeze a few more in the initial batch. I'm now close enough with the solvers that it's worth pushing through the rest of them. I'll let you know when it's ready again.

lmalenfant:
Thanks! Sounds good, I pulled the code locally so I could review it better and I will make some general comments to start with but leave the overall review until you have completed the solvers.

matlab/vts_wrapper/vts_mc_demo.m
```diff
@@ -8,7 +8,7 @@
 startup();
 
 % ======================================================================= %
-%% Example 1: run a simple Monte Carlo simulation with 1000 photons
+%% Example 01: run a simple Monte Carlo simulation with 1000 photons
```
lmalenfant, line 11: Can we either remove the colon here or add it in the Examples below?

src/Vts.Scripting.Test/ProgramTests.cs
lmalenfant, line 1: I think being able to run all the plots if good but I think in the test project we should mock the charting and verify that the values being passed are correct. We should have a unit test for each demo. If you would like me to look into this, I would be happy to help.

src/Vts.Scripting/ForwardSolvers/FS01_ROfRhoAndFtSingle.cs
```diff
@@ -0,0 +1,43 @@
+﻿using Vts.Common;
+using Vts.Modeling.ForwardSolvers;
+using Plotly.NET.CSharp;
+
+namespace Vts.Scripting.ForwardSolvers;
+
+/// <summary>
+/// Class using the Vts.dll library to demonstrate predicting reflectance as a function of rho and time frequency
+/// </summary>
+public class FS01_ROfRhoAndFtSingle : IDemoScript
```
lmalenfant, line 10: Now that these are classes, I think we need to revisit the names, these do not conform to any naming convention. I understand the need for the numbering to match the MATLAB numbers but we have to come up with a better solution.

dcuccia: Strongly disagree on this specific comment - these need to be script-first and match all the target destination files as the most helpful, most discoverable and browsable by Biophotonics  engineers (not developers). They're only classes to add testability, but never to be consumed directly by anyone else. Suggest we make them only visible to the test library as a compromise (`InternalsVisibleTo` attribute).

lmalenfant: I said revisit, at least let us discuss? What about Example01ROfRhoAndFtSingle - I would like to follow Pascal case and avoid underscores. The FS is redundant since they are in a folder named ForwardSolvers.

Yes I agree, these would be better as internal.

dcuccia: I think we are discussing. :) If the rationale for my position wasn't clear, I am happy to elaborate, but I am also fine with my opinion not being the winning argument. The proposed rationale provided as feedback for changing naming was because they were now classes. I don't think that's the best rationale in this specific case, which is why I disagreed. Here's a little more context for my stated opinion (which again, doesn't have to win, as it's ultimately the decision of the project lead):

These particular scripts existing in the Vts solution serve a specific, narrow purpose - to make testable code that would otherwise not be tested/compiled alongside the library that they reference, because they're destined to be in notebooks of various flavors. They are not intended to provide callable code from other parts of the codebase.

Therefore, I was proposing that we reconsider the "standard" C# naming conventions with respect to the consumer of this code ("the customer"), which is someone who will be opening a notebook, and running the examples with a single mouse click. While we do not (yet) have the final shape of what those example files will look like, my initial thinking was that they would all be in the same folder (right now in Matlab, they're all shoved in 2 files), which would mean that their names _would_ in fact need to disambiguate from each other across the different prefixes. The workflow of this user, in my view would be:

```pwsh
git clone https://github.com/VirtualPhotonics/VTS.git
cd VTS
python3 FS01_ + [TAB] to complete the name, FS01_ROfRhoAndFtSingle
```
This is what I meant by:

> the most helpful, most discoverable and browsable by Biophotonics engineers (not developers)

The scripts currently exist in folders only because I thought it would make them easier to refactor if our library dependencies change (e.g. if the solver demos in the future no longer needed to pull in a segregated "Vts.MonteCarlo.dll"). I could also see moving them all into a single "Examples" folder. In either case, I think a) imposing standard C# idioms onto code that is not intended to be called from outside C# consumers degrades the experience of the above use case without any obvious benefit (to me, at least - I'm open to feedback...), and b) that the underscores makes them more readable. In other words, I think it makes good sense to "work backwards" - to envision the final destination/goal, and design these names to match/meet that goal. (But if they don't match because they have to be different, ok).

My hope is that the names here can be shared with their intended destination to minimize any disconnect (e.g. `ForwardSolverExample01ROfRhoAndFtSinglecs` becoming `FS01_ROfRhoAndFtSingle.ipynb` seems confusing to me, and the latter is much more readable IMO).

Thoughts on any of the above?

dcuccia: Think I need to update the user story above, to disambiguate a .py script vs a .ipynb notebook. Command line use might not be as important as I was originally thinking. Will revise the stories tomorrow.

lmalenfant: That all makes sense. Do the names for Python have to be FS01_ etc? Do you think that someone using the Python scripts would be looking at the C# ones to need to make the correlation?

I think matching the MATLAB names as much as possible is good because those have been around for a while.

I think a good way around this would be to have Vts.Scripting as a separate repo also, then it is not putting code that does not conform to naming in the same project as Vts. This will also help the tests because we can leave them as is.

Do you think repo names like:
Vts.Scripting.CSharp
Vts.Scripting.Python 
would be good?

dcuccia: User stories are helpful :) I think I've talked myself around...

The likely workflow of the .NET Polyglot Notebook or Jupyter user would be:

```cmd
git clone https://github.com/VirtualPhotonics/VTS.git
cd VTS
code .
```

The last line opens the folder for visual browsing within VS Code. They'd then interact with the examples directly. If it's primarily a visual organization system, then it would be best to break things up into folders. 

Would `ForwardSovlers\Demo01ROfRhoAndFtSingle` and similar work?

> I think a good way around this would be to have Vts.Scripting as a separate repo also

`Vts.Scripting` was by design intentionally in the main repo so it could be compiled alongside the main code as developers do their work, and tested as part of the CI. Then, the plan was to create separate repos with notebooks (that used Nuget to pull in `Vts.dll`), which would be _mostly_ copy-pastable into .ipynb notebooks. I was thinking the naming might be something along the lines of :

vts-demos-python-notebooks
vts-demos-matlab
vts-demos-dotnet-notebooks

How's that sound? Long living "base" prefix `vts-demos` and then we can grow over time. Say, we want "raw" Python and JavaScript demos too, then we can add:

vts-demos-python
vts-demos-javascript
vts-demos-rust
vts-demos-go
vts-demos-swift
...

That work for you?

dcuccia: Also, I've now updated the 28 functions to make plotly chart rendering optional, and now skip it in the Vts.Scripting.Test tests.  Most viable alternative to mocking Plotly, which would require building undesired layers of indirection and hamper the intended purpose, which is to expose directly the full plotly capabilities/options to the script consumers. I spent a couple hours on mocking alternatives, and it's just not feasible based on the way Plotly.NET works, and what our intention is for users. I can explain more in-person if anyone's interested.

lmalenfant: `Would ForwardSovlers\Demo01ROfRhoAndFtSingle and similar work?`

Yes! I like that and if we name it correctly then it's ok to stay in the main repo. 

I'm good with the naming for the other repos because hyphens are the standard for those languages.

I think we went through a similar process with unit testing for Plotly and came to same conclusion so I like that we can skip the plotting in the tests.

lmalenfant:
@dcuccia I converted this to a draft, I think that will stop the email notifications until it is ready. I merged my branch with the NSubstitute changes so you will need to merge in master (The button above "Update branch" should be able to do that).

When you are ready for review, you can click the "Ready for Review" button.

If I'm telling you things you already know, please disregard :)

dcuccia:
Hi @lmalenfant, so sorry for the spam. I actually just finished the demos, so I'm marking it as ready for review now. In addition to a general code review, it would be great if @hayakawa16 could run some or all of the demos to spot-check my work. There was a lot of copy-paste, and I tried to go back and review, but it's a lot of stuff!

dcuccia:
I'd add one more thing: there's a lot to explore with Plotly, and we'll get a chance to dial in the visualizations, but hopefully (beyond a tweak here/there) that won't inhibit us from closing this initial effort. There is one thing in particular I haven't figured out, which is how to keep colorbars from overlapping. That's, and how to do better aspect ratio layout for multiple charts when combined. My guess is that as we go to Python demos, we'll learn what we _really_ want, and then have an opportunity to back-port those styles, etc to the C# side.

matlab/vts_wrapper/vts_solver_demo.m
```diff
@@ -8,7 +8,7 @@
 
 startup();
 
-%% Example ROfRhoAndFt
+%% Example 01 ROfRhoAndFt
```
lmalenfant, line 11: Add a colon to match above?

matlab/vts_wrapper/vts_solver_demo.m
```diff
@@ -343,14 +343,14 @@
...
-scatterer.vol_frac =  0.5;
+scatterer.vol_frac =  0.01;
 
 % % or 
 % scatterer.Type = 'Mie';
 % scatterer.radius =  0.5;
 % scatterer.n =       1.4;
 % scatterer.nMedium = 1.0;
-% scatterer.vol_frac = 0.5;
+% scatterer.vol_frac = 0.001;
```
lmalenfant, line 353: Just curious, why did you modify some of the values in the MATLAB?

src/Vts.Scripting.Test/Vts.Scripting.Test.csproj
```diff
@@ -0,0 +1,26 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.6.0" />
+    <PackageReference Include="NUnit" Version="3.13.3" />
+    <PackageReference Include="NUnit3TestAdapter" Version="4.4.2" />
+    <PackageReference Include="NUnit.Analyzers" Version="3.6.1" />
+	<PackageReference Include="NSubstitute" Version="5.0.0" />
+    <PackageReference Include="coverlet.collector" Version="3.2.0" />
+    <PackageReference Include="Typemock.Isolator" Version="9.2.3" />
```
lmalenfant, line 19: This dependency is not being used.

src/Vts.Scripting.Test/Vts.Scripting.Test.csproj
```diff
@@ -0,0 +1,26 @@
...
+    <IsPackable>false</IsPackable>
+    <IsTestProject>true</IsTestProject>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.6.0" />
+    <PackageReference Include="NUnit" Version="3.13.3" />
+    <PackageReference Include="NUnit3TestAdapter" Version="4.4.2" />
+    <PackageReference Include="NUnit.Analyzers" Version="3.6.1" />
+	<PackageReference Include="NSubstitute" Version="5.0.0" />
```
lmalenfant, line 17: This dependency is not being used.

src/Vts.Scripting/ForwardSolvers/Demo02ROfFxAndFtMulti.cs
```diff
@@ -0,0 +1,60 @@
...
+        var realAndImaginaryCharts = new[]
+        {
+            Chart.Combine(fxs.Select((fx, fxi) =>
+                LineChart(fts, rOfFxAndFt[fxi].Select(r => r.Real).ToArray(), xLabel, yLabel: $"R(ft) (real) [unitless]", title: $"real@fx={fx:F3}"))),
+            Chart.Combine(fxs.Select((fx, fxi) =>
+                LineChart(fts, rOfFxAndFt[fxi].Select(r => r.Imaginary).ToArray(), xLabel, yLabel: $"R(ft) (imag) [unitless]", title: $"imag@fx={fx:F3}")))
+        };
+        var grid2 = Chart.Grid(realAndImaginaryCharts, nRows: 2, nCols: 1, Pattern: Plotly.NET.StyleParam.LayoutGridPattern.Coupled);
+
+        if (showPlots)
```
lmalenfant, line 54: Invert if to reduce nesting.

src/Vts.Scripting/ForwardSolvers/Demo03FluenceOfRhoAndZ.cs
```diff
@@ -0,0 +1,71 @@
+﻿using Vts.Common;
+using Vts.Modeling.ForwardSolvers;
+using Vts.SpectralMapping;
+using Plotly.NET;
+using System;
```
lmalenfant, line 5: unused using.

src/Vts.Scripting/ForwardSolvers/Demo04FluenceOfRhoAndZAndFt.cs
```diff
@@ -0,0 +1,74 @@
+﻿using Vts.Common;
+using Vts.Modeling.ForwardSolvers;
+using Plotly.NET.CSharp;
+using System;
```
lmalenfant, line 4: Another unused using, maybe remove these globally in the scripting project (to avoid many file changes in case we still have them in VTS).

I think when you have implicit usings set to true, you no longer need to include these main usings.

src/Vts.Scripting/ForwardSolvers/Demo04FluenceOfRhoAndZAndFt.cs
```diff
@@ -0,0 +1,74 @@
...
+        var solver = new PointSourceSDAForwardSolver();
+        var op = new OpticalProperties(mua: 0.01, musp: 1, g: 0.8, n: 1.4);
+        var rhos = new DoubleRange(start: 0.1, stop: 19.9, number: 100).AsEnumerable().ToArray(); // range of s-d separations in mm
+        var zs = new DoubleRange(start: 0.1, stop: 19.9, number: 100).AsEnumerable().ToArray(); // range of depths in mm
+        var fts = new DoubleRange(start: 0, stop: 1, number: 2).AsEnumerable().ToArray(); // range of time frequencies in GHz
+
+        // predict the tissue's fluence(rho, z, ft) for the given optical properties 
+        var fluenceOfRhoAndZAndFt = solver.FluenceOfRhoAndZAndFt(new[]{ op }, rhos, zs, fts);
+
+        // collecte the steady-state (CW) and frequency domain fluence maps for plotting
```
lmalenfant, line 34: Typo collecte :)

src/Vts.Scripting/ForwardSolvers/Demo04FluenceOfRhoAndZAndFt.cs
```diff
@@ -0,0 +1,74 @@
...
+        var cwFluenceOfRhoAndZ = fluenceOfRhoAndZAndFt.TakeEveryNth(n: fts.Length, skip: ft1);
+        var fdFluenceOfRhoAndZ = fluenceOfRhoAndZAndFt.TakeEveryNth(n: fts.Length, skip: ft2);
+
+        var imageSize = rhos.Length * zs.Length;
+        var allRhos = rhos.Select(rho => -rho).Reverse().Concat(rhos).ToArray(); // duplicate for -rho to make symmetric
+        // Plot the CW fluence: log(fluence(rho, z, ft=0GHz)) 
+        var cwFluenceRowsToPlot = cwFluenceOfRhoAndZ
+            .Select(fluence => Math.Log(fluence.Magnitude)) // take log for visualization purposes
+            .Chunk(zs.Length); // break the heatmap into rows (inner dimension is zs)        
+        var allCwFluenceRowsToPlot = cwFluenceRowsToPlot.Reverse().Concat(cwFluenceRowsToPlot).ToArray(); // duplicate for -rho to make symmetric
```
lmalenfant, line 46: Possible multiple enumeration

src/Vts.Scripting/ForwardSolvers/Demo04FluenceOfRhoAndZAndFt.cs
```diff
@@ -0,0 +1,74 @@
...
+        // predict the tissue's fluence(rho, z, ft) for the given optical properties 
+        var fluenceOfRhoAndZAndFt = solver.FluenceOfRhoAndZAndFt(new[]{ op }, rhos, zs, fts);
+
+        // collecte the steady-state (CW) and frequency domain fluence maps for plotting
+        // (innermost axis is ft, so we need to take every 2nd starting at index 0)
+        var (ft1, ft2) = (0, 1); // indices of the two time frequencies, ft, to plot
+        var cwFluenceOfRhoAndZ = fluenceOfRhoAndZAndFt.TakeEveryNth(n: fts.Length, skip: ft1);
+        var fdFluenceOfRhoAndZ = fluenceOfRhoAndZAndFt.TakeEveryNth(n: fts.Length, skip: ft2);
+
+        var imageSize = rhos.Length * zs.Length;
```
lmalenfant, line 40: not used

src/Vts.Scripting/ForwardSolvers/Demo04FluenceOfRhoAndZAndFt.cs
```diff
@@ -0,0 +1,74 @@
...
+        var allRhos = rhos.Select(rho => -rho).Reverse().Concat(rhos).ToArray(); // duplicate for -rho to make symmetric
+        // Plot the CW fluence: log(fluence(rho, z, ft=0GHz)) 
+        var cwFluenceRowsToPlot = cwFluenceOfRhoAndZ
+            .Select(fluence => Math.Log(fluence.Magnitude)) // take log for visualization purposes
+            .Chunk(zs.Length); // break the heatmap into rows (inner dimension is zs)        
+        var allCwFluenceRowsToPlot = cwFluenceRowsToPlot.Reverse().Concat(cwFluenceRowsToPlot).ToArray(); // duplicate for -rho to make symmetric
+        var cwFluenceChart = Heatmap(values: allCwFluenceRowsToPlot, x: allRhos, y: zs, xLabel: "ρ", yLabel: "z", title: "log(Φ(ρ, z, ft=0Ghz))");
+
+        // Plot the frequency-domain fluence amplitude at 1GHz: log(fluence(rho, z, ft=0GHz)) 
+        var fdFluenceRowsToPlot = fdFluenceOfRhoAndZ
```
lmalenfant, line 50: Possible multiple enumeration

src/Vts.Scripting/ForwardSolvers/Demo04FluenceOfRhoAndZAndFt.cs
```diff
@@ -0,0 +1,74 @@
...
+        // collecte the steady-state (CW) and frequency domain fluence maps for plotting
+        // (innermost axis is ft, so we need to take every 2nd starting at index 0)
+        var (ft1, ft2) = (0, 1); // indices of the two time frequencies, ft, to plot
+        var cwFluenceOfRhoAndZ = fluenceOfRhoAndZAndFt.TakeEveryNth(n: fts.Length, skip: ft1);
+        var fdFluenceOfRhoAndZ = fluenceOfRhoAndZAndFt.TakeEveryNth(n: fts.Length, skip: ft2);
+
+        var imageSize = rhos.Length * zs.Length;
+        var allRhos = rhos.Select(rho => -rho).Reverse().Concat(rhos).ToArray(); // duplicate for -rho to make symmetric
+        // Plot the CW fluence: log(fluence(rho, z, ft=0GHz)) 
+        var cwFluenceRowsToPlot = cwFluenceOfRhoAndZ
```
lmalenfant, line 43: Possible multiple enumeration

src/Vts.Scripting/ForwardSolvers/Demo03FluenceOfRhoAndZ.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        // Plot the log(fluence(rho, z)) for the last wavelength
+        var wvi = wavelengths.Length - 1; // index of last wavelength (1000nm)
+        var imageSize = rhos.Length * zs.Length;
+        var allRhos = rhos.Select(rho => -rho).Reverse().Concat(rhos).ToArray(); // duplicate for -rho to make symmetric
+        var fluenceRowsToPlot = fluenceOfRhoAndZ            
+            .Skip(wvi * imageSize) // skip to the last wavelength
+            .Select(fluence => Math.Log(fluence)) // take log for visualization purposes
+            .Chunk(zs.Length); // break the heatmap into rows (inner dimension is zs)        
+        var fluenceDataToPlot = fluenceRowsToPlot.Reverse().Concat(fluenceRowsToPlot).ToArray(); // duplicate for -rho to make symmetric
```
lmalenfant, line 62: possible multiple enumeration

src/Vts.Scripting/ForwardSolvers/Demo05PHDOfRhoAndZ.cs
```diff
@@ -0,0 +1,52 @@
...
+using Plotly.NET.CSharp;
+using Vts.Factories;
+
+namespace Vts.Scripting.ForwardSolvers;
+
+/// <summary>
+/// Class using the Vts.dll library to demonstrate predicting photon hitting density (PHD)
+/// as a function of radial extent and depth at a given set of optical properties and source-detector separation
+/// </summary>
+internal class Demo05PHDOfRhoAndZ : IDemoScript
```
lmalenfant, line 12: This name "should" be Demo05PhdOfRhoAndZ

Having said that I do realize that this is wrong in many places in our solution, my preference would be to change it here but I do understand if you want to keep it this way.

lmalenfant:
@dcuccia the majority of my comments are coming from Resharper suggestions, I think I remember you said you had Resharper so would you like to go though and fix these before our review?

If you do not have Resharper, I would be happy to make these changes for you.

matlab/vts_wrapper/vts_solver_demo.m
dcuccia: Some of the Matlab values were bad - didn't mimic tissue at all. They're consistent with the C# side.

src/Vts.Scripting/ForwardSolvers/Demo02ROfFxAndFtMulti.cs
dcuccia: I intended this one, because I want to make it simple for new developers to edit and append to the script. Just because the "demo" class exits finally in this version, doesn't mean someone might not want to do a follow-on computation afterward. If they did, this ReSharper suggestion would go away. So, it's defensive coding for novice developer happiness.

src/Vts.Scripting/ForwardSolvers/Demo04FluenceOfRhoAndZAndFt.cs
dcuccia: That's a good idea, thanks. In fact, it's likely the notebooks will have a "header" that will bring in all needed usings, and the Implcit usings + GlobalUsings will mimic that best.

src/Vts.Scripting.Test/Vts.Scripting.Test.csproj
dcuccia: Oh, sorry that stayed in there. Will keep NSubstitute, but remove Typemock. FWIW, that was from my trying to implement a "shim" for static Plotly methods, and unfortunately it is not free and requires specific Visual Studio tooling. Same thing with Microsoft's "Fakes" - requires VS Ultimate. I did discover Pose, but per [this thread](https://github.com/tonerdo/pose/issues/63), it's not currently well-supported.

src/Vts.Scripting/ForwardSolvers/Demo05PHDOfRhoAndZ.cs
dcuccia: It's kind of a question of "glanceable comprehension" by a photonics domain expert, vs correctness from a pure C# developer perspective. The function that generates it is, itself, `GetPHD` (which was intentional when we created it). Same thing with CAW vs DAW, etc.

lmalenfant: 05PHDO is what I see when I glance. I'm not sure the answer to this one. I had a tough time with this in the beginning for things like PDF and API but now I abide by it because these are know acronyms and the Pascal case defines the separation. Honestly, here I would be ok with either way.

@hayakawa16 and I have been changing these in the code to follow C# standards. At the end of the day if you are programming in a language is is good to abide by the rules of the language. For example wouldn't this be in lower case delimited with underscores in Python?

matlab/vts_wrapper/vts_mc_demo.m
dcuccia: the MC demos all had it, but I've updated the other scripts to match as well

src/Vts.Scripting/ForwardSolvers/Demo04FluenceOfRhoAndZAndFt.cs
dcuccia: Thanks, got all of these I think, except where that's intentional (for the `Reverse` calls)

dcuccia:
Changes applied, except where noted above in conversations.

src/Vts.Scripting/ForwardSolvers/Demo05PHDOfRhoAndZ.cs
dcuccia: I see what you mean (one reason I had an underscore in there originally). Either is fine with me, "ideomatic C#" and "ideomatic Python" probably make sense. Yes, Python often uses underscores in names, sometimes `PragmaticCamelCase_WithAbbreviationsAndUnderscoresLike_RGB()` for class members.

dcuccia: Here are all the changes I've applied, but haven't yet committed (all make sense?):

`Demo05PHDOfRhoAndZ` -> `Demo05PhdOfRhoAndZ`
`Demo07PHDOfRhoAndZTwoLayer` -> `Demo07PhdOfRhoAndZTwoLayer`
`Demo02DAWvsCAW` -> `Demo02DawVsCaw`
`Demo04N1000vsN100` -> `Demo04N1000VsN100`
`Demo06pMCPostProcessor` -> `Demo06PmcPostProcessor`
`Demo07pMCInversion` -> `Demo07PmcInversion`

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
lmalenfant, line 1: I think you need to add this folder to the gitignore

src/Vts.Scripting/ForwardSolvers/Demo01ROfRhoAndFtSingle.cs
```diff
@@ -0,0 +1,44 @@
...
+        // PointSourceSDA,DistributedGaussianSourceSDA, DistributedPointSourceSDA,
+        // MonteCarlo(basic scaled), Nurbs(scaled with smoothing and adaptive binning)
+        var solver = new PointSourceSDAForwardSolver();
+        var op = new OpticalProperties(mua: 0.01, musp: 1.2, g: 0.8, n: 1.4);
+        var rho = 10; // s-d separation, in mm
+        var fts = new DoubleRange(start: 0, stop: 0.5, number: 51).AsEnumerable().ToArray(); // range of temporal frequencies in GHz
+
+        // predict the temporal frequency response at the specified optical properties and s-d separation
+        var rOfFt = solver.ROfRhoAndFt(op, rho, fts);
+        var magnitudeScale = 1E4;
```
lmalenfant, line 25: Convert to constant

src/Vts.Scripting/ForwardSolvers/Demo01ROfRhoAndFtSingle.cs
```diff
@@ -0,0 +1,44 @@
...
+        var rOfFtAmplitude = rOfFt.Select(r => r.Magnitude * magnitudeScale).ToArray();
+        var rOfFtPhase = rOfFt.Select(r => -r.Phase).ToArray();
+
+        //%% Example ROfRhoAndFt
+        //% Evaluate reflectance as a function of rho and temporal-frequency with one 
+        var xLabel = "time frequency [GHz]";
```
lmalenfant, line 31: Convert to constant

src/Vts.Scripting/ForwardSolvers/Demo01ROfRhoAndFtSingle.cs
```diff
@@ -0,0 +1,44 @@
...
+    public static void RunDemo(bool showPlots = true)
+    {
+        // Example 01: predict R(rho, ft) based on a standard diffusion approximation solution to the time-dependent RTE
+
+        // Solver type options:
+        // PointSourceSDA,DistributedGaussianSourceSDA, DistributedPointSourceSDA,
+        // MonteCarlo(basic scaled), Nurbs(scaled with smoothing and adaptive binning)
+        var solver = new PointSourceSDAForwardSolver();
+        var op = new OpticalProperties(mua: 0.01, musp: 1.2, g: 0.8, n: 1.4);
+        var rho = 10; // s-d separation, in mm
```
lmalenfant, line 20: Convert to constant

src/Vts.Scripting/ForwardSolvers/Demo02ROfFxAndFtMulti.cs
```diff
@@ -0,0 +1,56 @@
...
+internal class Demo02ROfFxAndFtMulti : IDemoScript
+{
+    /// <summary>
+    /// Sample script to demonstrate this class' stated purpose
+    /// </summary>
+    public static void RunDemo(bool showPlots = true)
+    {
+        // Example 02: predict R(fx, ft) based on a standard diffusion approximation solution to the time-dependent RTE
+        // Note:
+        //     - this example computes and displays R(ft) at mutiple spatial frequencies, to demonstrate that multiplexing functionality
```
lmalenfant, line 15: typo multiple

src/Vts.Scripting/ForwardSolvers/Demo02ROfFxAndFtMulti.cs
```diff
@@ -0,0 +1,56 @@
...
+        //     - updating the fx range to emit a single value will isloate to the frequency desired
```
lmalenfant, line 16: Typo isolate

lmalenfant:
@dcuccia I am still seeing issues in the files like making variables global, typos and unnecessary string interpolation, I'm not sure if these are caught by Resharper or SonarLint (which is another good plugin that we use). Would you be able to take another pass?

At any time, I am more than willing to dig in, fixing is easier than pointing out all the issues.

src/Vts.Scripting/ForwardSolvers/Demo05PHDOfRhoAndZ.cs
lmalenfant: I like those, do you want to fix the OP -> Op also? :)

dcuccia:
The constant and interpolation comments are overkill and not important for this use case. Will update the typos.

src/Vts.Scripting/ForwardSolvers/Demo05PHDOfRhoAndZ.cs
dcuccia: No, it's a two letter acronym and not an abbreviation

https://stackoverflow.com/a/596070

lmalenfant: That post is 14 years old, Microsoft call it Id now and Db etc.

Maybe write out OpticalProperties then like we do in the code?

dcuccia: > By convention, two-letter acronyms use all uppercase letters, and acronyms of three or more characters use Pascal casing. The following examples use this naming convention: 'DB', 'CR', 'Cpa', and 'Ecma'. The following examples violate the convention: 'Io', 'XML', and 'DoD', and for non-parameter names, 'xp' and 'cpl'. 'ID' is special-cased to cause a violation of this rule. 'Id' is not an acronym but is an abbreviation for 'identification'.

https://learn.microsoft.com/en-us/visualstudio/code-quality/ca1709?view=vs-2022

dcuccia: It's also true that abbreviations are discouraged. If we think the extra name length isn't ridiculous, I'm ok spelling them out, just seemed long.

lmalenfant: That's a better article, makes sense.

lmalenfant: > It's also true that abbreviations are discouraged. If we think the extra name length isn't ridiculous, I'm ok spelling them out, just seemed long.

Your call, I like writing things out also but I can live with this if it's comprehendible.

.vscode/settings.json
dcuccia: This one was intentional. It let's us share settings as a team, and this setting in particular enables an out-of-the-box experience for the new VS Code feature ([C# Dev Kit](https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.csdevkit)) that allows developing/building/running at the solution level. This is particularly relevant to this branch - now that there's C# Dev Kit, there's nothing that prevents a Linux or Mac user from installing VS Code and running these script samples directly in the solution. If we wanted to discourage modification of the file, we could add it to our .gitignore. More helpful guidance here, including discussion of suggested default VS code extensions as well, like C# Dev Kit, Python/Jupyter, Matlab syntax highlighting/running):

https://bobbyhadz.com/blog/what-is-vscode-folder#should-you-commit-your-vscode-directory-to-source-control

dcuccia:
Installing ReSharper to see what else might be work fixing, and will update when I'm done.

src/Vts.Scripting/ForwardSolvers/Demo05PHDOfRhoAndZ.cs
dcuccia: Slept on it - thinking "OpProp" is a good balance to strike

dcuccia:
Ok, all done. @hayakawa16 let me know when you might have a chance to glance through the samples for correctness (labels, axes, etc). No hurry!

hayakawa16:
Hi, I just pulled latest.  I've been following along with the code cleanup but haven't run demos in a while.  I set Vts.Scripting to the startup project, then ran with green right arrow.  Only one demo ran and I *think* it is ForwardSolvers Demo18. This looks like the x-axis label should be wavelength, not rho, and likewise y-axis should be log(R(lambda)).
How can I run individual demo class so I know what I'm running and so I can inform you as to if I see error in axes labels?

dcuccia:
You can change the name of the demo to run in the Program.cs file of the script project, or change the constructor to call the RunAll... Function for forward solvers, mc, short course, and then step through them all sequentially

hayakawa16:
Thanks.  I have the following questions:
1) Demo04FluenceOfRhoAndZAndFt: The title is "modulation" rather than "fluence" and there are 2 extra right ")"
2) Demo05PhdOfRhoAndZ: I don't really see "banana" even with a very major log scale.  If you pull up WPF GUI and set same settings, a very different plot is shown.  First, the isotropic point source under rho=0 is not seen, it looks more like a linear decay of that source under rho=0. and there is no banana halo, is this due to the color scheme?
3) Demo07PhdOfRhoAndZTwoLayer: I see the "banana" here!  Although on this and Demo06FluenceOfRhoAndZTwoLayer there is a lighter line at about z=0.7mm, which is not the top layer thickness=5mm.  I brought up the WPF GUI and don't see that lighter line.  I wonder what is causing it?  
4) Demo08AbsorbedEnergyOfRhoAndZ: I don't think the units are correct.  The "mm-3" is correct, however is "mW-1" suppose to mean inverse milli-Watts?  If so, I don't think this is correct.  I would believe "W" watts in numerator, not sure where the "mW" is coming from.
5) Demo10ROfRhoAndTimeMulti: the MATLAB script defines a single rho and this script defines 19, not a problem but on the 2 plots it is hard to see any difference between all of the cases, possibly because it is on log scale?  Possibly fewer rho, and/or smaller max t, and/or regular scale rather than log?  Also on the units there is a "s-1" which I think should be "ns-1".
6) Demo11ROfFxAndTime and Demo12ROfFxSingle: The former has the y-axis label with absolute value signs, however the latter does not have these signs.  In both cases, the plot is of amplitude correct?
7) Demo16ROfFxMultiPowerLaw: interesting plot, at first didn't know what "A" was but then looked at code.  Nice!
8) Demo18ROfRhoMultiOpPropInversion: x-axis should be "wavelength" and y-axis "log(R(lambda))".
9) Demo21ROfRhoAndFtTwoLayerMultiOpProp: same comment about units with respect to time, y-axis units should be "ns-1" instead of "s-1"

Great work!  A lot of effort here.  Thank you.  I will work on the demos in MonteCarlo and ShortCourse next.

dcuccia:
Thanks Carole! I will take a look at all of these, might take me a day or so, depending on how my Sunday goes.

src/Vts.Scripting/ForwardSolvers/Demo05PHDOfRhoAndZ.cs
lmalenfant: That's funny, you had just convinced me that 2 letter acronyms were acceptable as uppercase, now I am ok with IO and DB. I do think these are more readable the way you have them though.

hayakawa16:
I should mention that my suggested modifications are just suggestions.  For example on 8. above I think you've been using "wavelength" as the axis label, so the y-axis of "log(R(wavelength))" makes more sense.  I think if it is consistent across plots then it would be good.

src/Vts.Scripting/ForwardSolvers/Demo03FluenceOfRhoAndZ.cs
```diff
@@ -0,0 +1,66 @@
...
+        var chromophores = new IChromophoreAbsorber[]
+        {
+            new ChromophoreAbsorber(ChromophoreType.HbO2, 70), // molar concentration
+            new ChromophoreAbsorber(ChromophoreType.Hb, 30), // molar concentration
+            new ChromophoreAbsorber(ChromophoreType.H2O, 0.8), // fractional concentration
+        };
+
+        // construct a scatterer
+        var scatterer = new PowerLawScatterer(a: 1.2, b: 1.42);
+        // or: var scatterer = new IntralipidScatterer(volumeFraction: 0.5);
```
lmalenfant, line 37: Commented out code causes code smells but I understand the need for this, one thing I have discovered is removing the semi-colon will remove the "smell". 

If you want to leave it as is though, I understand.

src/Vts.Scripting/ForwardSolvers/Demo06FluenceOfRhoAndZTwoLayer.cs
```diff
@@ -0,0 +1,67 @@
...
+                zRange: new DoubleRange(topLayerThickness, double.PositiveInfinity, 2),
+                regionOP: new OpticalProperties(mua: 0.1, musp: 1, g: 0.8, n: 1.4)
+            )
+        };
+        var rhos = new DoubleRange(start: 0.1, stop: 19.9, number: 100).AsEnumerable().ToArray(); // range of s-d separations in mm
+        var zs = new DoubleRange(start: 0.1, stop: 19.9, number: 100).AsEnumerable().ToArray(); // range of depths in mm
+
+        // predict the tissue's fluence(rho, z) for the given optical properties 
+
+        //var fluenceOfRhoAndZ = solver.FluenceOfRhoAndZ(new[] { opRegions }, rhos, zs );
```
lmalenfant, line 40: Is this commented code intentional? If so could we explain?

src/Vts.Scripting/ForwardSolvers/Demo07PHDOfRhoAndZTwoLayer.cs
```diff
@@ -0,0 +1,71 @@
...
+            ( 
+                zRange: new DoubleRange(topLayerThickness, double.PositiveInfinity, 2),
+                regionOP: new OpticalProperties(mua: 0.1, musp: 1, g: 0.8, n: 1.4)
+            )
+        };
+        var rhos = new DoubleRange(start: 0.1, stop: 19.9, number: 100).AsEnumerable().ToArray(); // range of s-d separations in mm
+        var zs = new DoubleRange(start: 0.1, stop: 19.9, number: 100).AsEnumerable().ToArray(); // range of depths in mm
+
+        // predict the tissue's fluence(rho, z) for the given optical properties 
+        //var fluenceOfRhoAndZ = solver.FluenceOfRhoAndZ(new[] { opRegions }, rhos, zs );
```
lmalenfant, line 39: Is this commented code intentional? If so could we explain?

hayakawa16:
For the MonteCarlo folder, all looks terrific!
For the ShortCourse folder:
1) Demo01APhotonCountWithFluence: you mentioned this already about the overlapping colorbar.    Otherwise, these are great plots!
2) Demo01BAnalogVsDiscreteWithFluence: this puts out 8 tabs.  I happen to know these plots well from the short course so I know what they are.  This might be a big ask, but at the top where is says "log(Phi(rho,z))", could this be extended to list which random walk process is being used and how big N is?  Something like "log(Phi(rho,z)) Analog N=100" or something like that?
3) Demo02AnalogVsContinuousWithReflectance: on the top plot, the y-axis label should just be "log(R(rho)) [mm-2])" and then a length with red line = analog, blue line=CAW somewhere.  On bottom plot, I would prefer "Analog" instead of "AAW" because "AW" stands for "absorption weighting" and analog has no absorption weighting.

Otherwise, looking good!

src/Vts.Scripting/MonteCarlo/Demo06pMCPostProcessor.cs
```diff
@@ -0,0 +1,121 @@
...
+        //    - default source is a point source beam normally incident at the origin 
+
+        // create a SimulationInput object to define the simulation
+        var tissueInput = new MultiLayerTissueInput
+        {
+            Regions = new ITissueRegion[]
+                {
+                    new LayerTissueRegion(
+                        zRange: new(double.NegativeInfinity, 0),         // air "z" range
+                        op: new(mua: 0.0, musp: 1E-10, g: 1.0, n: 1.0)), // air optical properties
```
lmalenfant, line 25: I'm seeing an alert here to specify the type (and in several other places), was this intentional to keep the code more concise or align with Python?

Since you have labeled them, I don't think adding the type adds anything except showing which classes are being utilized.

src/Vts.Scripting/ShortCourse/Demo01APhotonCountWithFluence.cs
```diff
@@ -0,0 +1,124 @@
...
+        // compute the relative error (standard deviation / mean) for each simulation
+        var allRelativeErrors = numPhotonsArray.Select((numPhotons, npIdx) =>
+            allFluenceMeans[npIdx].Zip(allFluenceSecondMoments[npIdx], (mean, secondMoment) => 
+                Math.Sqrt((secondMoment - mean * mean) / numPhotons) / mean).ToArray()).ToArray();
+
+        // plot the results using Plotly.NET
+        var rhos = rhoRange.GetMidpoints();
+        var zs = zRange.GetMidpoints();
+        var allRhos = rhos.Select(rho => -rho).Reverse().Concat(rhos).ToArray(); // duplicate for -rho to make symmetric
+        var charts = numPhotonsArray.Select((numPhotons, npIdx) =>
```
lmalenfant, line 94: numPhotons is not used replace with _?

src/Vts.Scripting/ShortCourse/Demo01BAnalogVsDiscreteWithFluence.cs
```diff
@@ -0,0 +1,146 @@
...
+        var analogRelativeErrors = allRelativeErrors.Take(4).ToArray();
+        var discreteRelativeErrors = allRelativeErrors.Skip(4).ToArray();
+        var relativeErrorDifference = analogRelativeErrors.Zip(discreteRelativeErrors, (analog, discrete) =>
+                analog.Zip(discrete, (a, d) => a - d).ToArray()).ToArray();
+
+        // plot the results using Plotly.NET
+        var rhos = rhoRange.GetMidpoints();
+        var zs = zRange.GetMidpoints();
+        var allRhos = rhos.Select(rho => -rho).Reverse().Concat(rhos).ToArray(); // duplicate for -rho to make symmetric
+        var fluenceAndRelativeErrorCharts = simulationTuples.Select((tuple, tupleIdx) =>
```
lmalenfant, line 108: Tuple is not used, replace with _

src/Vts.Scripting/ShortCourse/Demo01BAnalogVsDiscreteWithFluence.cs
```diff
@@ -0,0 +1,146 @@
...
+                .ToArray();
+            var relativeErrorDataToPlot = relativeErrorRowsToPlot.Reverse().Concat(relativeErrorRowsToPlot).ToArray(); // duplicate for -rho to make symmetric
+            var relativeErrorMap = Heatmap(values: relativeErrorDataToPlot, x: allRhos, y: zs,
+                xLabel: "ρ [mm]", yLabel: "z [mm]", title: $"error(ρ, z)");
+
+            var combined = Chart.Grid(new[]{ fluenceMap, relativeErrorMap }, nRows: 2, nCols: 1, Pattern: Plotly.NET.StyleParam.LayoutGridPattern.Coupled);
+
+            return combined;
+        });
+        var analogVsDiscreteRelativeErrorCharts = relativeErrorDifference.Select((difference, differenceIdx) =>
```
lmalenfant, line 129: difference is not used replace with _

src/Vts.Scripting/ShortCourse/Demo02AnalogVsContinuousWithReflectance.cs
```diff
@@ -0,0 +1,108 @@
...
+        var allSimulationOutputs = MonteCarloSimulation.RunAll(allSimulations);
+
+        // collect all the results in helpful arrays
+        var allReflectanceDetectors = allSimulationOutputs.Select(detectorResult => 
+            (ROfRhoDetector)detectorResult.ResultsDictionary["ROfRho"]).ToArray();
+        var allReflectanceMeans = allReflectanceDetectors.Select(detector => detector.Mean).ToArray();
+        var allReflectanceSecondMoments = allReflectanceDetectors.Select(detector => detector.SecondMoment).ToArray();
+
+        // compute the relative error (standard deviation / mean) for each simulation
+        var allRelativeErrors = absorptionWeightingTypes.Select((awt, awtIdx) =>
```
lmalenfant, line 83: awt is not used, replace with _?

src/Vts.Scripting/IDemoScript.cs
```diff
@@ -0,0 +1,12 @@
...
+
+/// <summary>
+/// Interface to constrain all demo scripts to a uniform signature
+/// </summary>
+interface IDemoScript
+{
+    /// <summary>
+    /// The one required static method for all demo scripts
+    /// </summary>
+    static abstract void RunDemo(bool showPlots = true);
```
lmalenfant, line 11: abstract is not supported yet in this version of .NET

src/Vts.Scripting/Program.cs
```diff
@@ -0,0 +1,73 @@
+﻿[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("Vts.Scripting.Test")]
+namespace Vts.Scripting;
+
+/// <summary>
+/// Top program used to execute the desired demo script(s)
+/// </summary>
+public class Program
+{
```
lmalenfant, line 8: Have you considered having the minimal program.cs file and moving the run demos into their own class?

It is so much easier to read for a newbie than having the Main.

dcuccia: I can do that!

src/Vts.Scripting/ShortCourse/Demo01APhotonCountWithFluence.cs
dcuccia: This (and all similar examples) is an optimization that doesn't help someone using this code as a jumping off point. In the general case, they'll want access to the main variable they're looping over.

src/Vts.Scripting/ShortCourse/Demo01BAnalogVsDiscreteWithFluence.cs
dcuccia: Same as above

src/Vts.Scripting/IDemoScript.cs
dcuccia: This is why I targeted .NET 7 (for this _consuming_ app only, not the published Vts library) - it's a way to add constraints on static methods, and generally improves things. If it's absolutely necessary to downgrade, that's fine. But any C# developer will have .NET 7 and I don't see a good reason to limit ourselves artificially when we're consuming the published/tested Vts.

src/Vts.Scripting/MonteCarlo/Demo06pMCPostProcessor.cs
dcuccia: What does an "alert" mean, in this case? Are there language features we can't use? Target typed properties remove duplication of info, and I thought perhaps easier to read, in cases like this where you're trying to fit a big idea on the screen.

lmalenfant:
@dcuccia this is looking good, none of my recent comments are critical to fix or even need to be fixed if that was your intention. Mainly pointing them out because I noticed them.

There are some of the local variable names that are a little ambiguous to me (like tIdx0p01), not sure if they could be named to be more clear what they are. Maybe they would be clear to someone who knows the subject better.

src/Vts.Scripting/ForwardSolvers/Demo07PHDOfRhoAndZTwoLayer.cs
dcuccia: Thanks for catching this. You know, I was having trouble with that line compiling, so I reverted to the `ComputationFactory.ComputeFluence` version, and was thinking to add a "todo" to fix/add the overload that would make that code work. But now, as I uncomment it...it works? :\ So I'll go with that instead and adjust the code as needed, accordingly.

src/Vts.Scripting/IDemoScript.cs
lmalenfant: You can add preview to the language in csproj to keep that in place.

We stick with the LTS versions so we don't have to upgrade as often.

src/Vts.Scripting/IDemoScript.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Vts.Scripting;
+
+/// <summary>
+/// Interface to constrain all demo scripts to a uniform signature
+/// </summary>
+interface IDemoScript
```
lmalenfant, line 6: Add internal modifier

src/Vts.Scripting/ForwardSolvers/Demo06FluenceOfRhoAndZTwoLayer.cs
dcuccia: Same comment as below

src/Vts.Scripting/MonteCarlo/Demo06pMCPostProcessor.cs
lmalenfant: I'm not sure if this is Resharper or SonarLint, sometimes it advises removing the type but in this case is suggests specifying it. Like I said though, I don't think adding it in this case makes it better but that's mainly because you have them labeled, without that I could see where the type name would be useful.

src/Vts.Scripting/ShortCourse/Demo01APhotonCountWithFluence.cs
lmalenfant: That's what I figured, just being thorough in my review :)

src/Vts.Scripting/MonteCarlo/Demo06pMCPostProcessor.cs
lmalenfant: I think this is SonarLint and it can be changed to always prefer code without the type so I will resolve this conversation because I think you should leave it the way it is.

src/Vts.Scripting/IDemoScript.cs
dcuccia: Thanks for all the add'l comments. :) Ok, sounds good. Maybe if we can get automated CI/CD working, I can change your mind. 😇

dcuccia:
Carole, I've tried to answer/address your questions inline below
> Thanks. I have the following questions:
> Demo04FluenceOfRhoAndZAndFt: The title is "modulation" rather than "fluence" and there are 2 extra right ")"

Yes, three plots: 1) Fluence at 0GHz, 2) Fluence at 1GHz, 3) the Modulation - the ratio of 1ghz to DC. That make sense?

> Demo05PhdOfRhoAndZ: I don't really see "banana" even with a very major log scale. If you pull up WPF GUI and set same settings, a very different plot is shown. First, the isotropic point source under rho=0 is not seen, it looks more like a linear decay of that source under rho=0. and there is no banana halo, is this due to the color scheme?

No, I'm not sure if it's just the colormap. I changed it to Hot, and got this:
PHD

But..AHA! It was a typo and computing `FluenceOfFxAndZ` :) That would do it! Works now:
phd fixed

Thanks a lot for catching this.

> Demo07PhdOfRhoAndZTwoLayer: I see the "banana" here! Although on this and Demo06FluenceOfRhoAndZTwoLayer there is a lighter line at about z=0.7mm, which is not the top layer thickness=5mm. I brought up the WPF GUI and don't see that lighter line. I wonder what is causing it?

I don't know either. It's pretty weird, but I don't have a good guess after beard scratching for a little bit.

> Demo08AbsorbedEnergyOfRhoAndZ: I don't think the units are correct. The "mm-3" is correct, however is "mW-1" suppose to mean inverse milli-Watts? If so, I don't think this is correct. I would believe "W" watts in numerator, not sure where the "mW" is coming from.

I mulled this one for a bit and thought it would need to be per unit light dose or power put into to the tissue, thinking the input would be in Watts or Joules. But, I guess it's just a "normalized" mm-3 (per depth, per surface area). Fluence is a value normalized to unit input - I guess that input could be a power (W) or energy dose (J)?  I always thought of it as a fluence rate - so W would be the right units for a non-normalized rate of Energy Absorption. I'll keep it normalized for sure. Would be interesting to have your and the team's input on that, though.

> Demo10ROfRhoAndTimeMulti: the MATLAB script defines a single rho and this script defines 19, not a problem but on the 2 plots it is hard to see any difference between all of the cases, possibly because it is on log scale? Possibly fewer rho, and/or smaller max t, and/or regular scale rather than log? Also on the units there is a "s-1" which I think should be "ns-1".

Good point, thanks. Might have just been copy-paste from the previous example to try to keep all rhos. I guess if you plot log-log, it looks like this:
image

...but I went ahead and simplified and just put on a linear axis:
image
Look ok?

> Demo11ROfFxAndTime and Demo12ROfFxSingle: The former has the y-axis label with absolute value signs, however the latter does not have these signs. In both cases, the plot is of amplitude correct?

Thanks for finding the inconsistency. Removed the absolute value signs - think that would probably just confuse people anyway (we can have phase shifts in the SFD, but we don't ever discuss those in the current demo, and probably only useful to use magnitude in that future case where we're differentiating from spatial phase).

> Demo16ROfFxMultiPowerLaw: interesting plot, at first didn't know what "A" was but then looked at code. Nice!

Thanks :)

> Demo18ROfRhoMultiOpPropInversion: x-axis should be "wavelength" and y-axis "log(R(lambda))".

Thanks!

> Demo21ROfRhoAndFtTwoLayerMultiOpProp: same comment about units with respect to time, y-axis units should be "ns-1" instead of "s-1"

Thanks, fixed that in 3 files

dcuccia:
@lmalenfant 
> There are some of the local variable names that are a little ambiguous to me (like tIdx0p01), not sure if they could be named to be more clear what they are. Maybe they would be clear to someone who knows the subject better.

Maybe I'm too close to it, but that's 0p123 is a relatively common way we computational people get around decimals not working well in variable names. Defer to @hayakawa16 ...

dcuccia:
@hayakawa16 

> For the ShortCourse folder:

> Demo01APhotonCountWithFluence: you mentioned this already about the overlapping colorbar. Otherwise, these are great plots!

Yeah, sorry - I'll spend some time on a bug branch trying to fix these. Too confusing for them to be separate, I thought.

> Demo01BAnalogVsDiscreteWithFluence: this puts out 8 tabs. I happen to know these plots well from the short course so I know what they are. This might be a big ask, but at the top where is says "log(Phi(rho,z))", could this be extended to list which random walk process is being used and how big N is? Something like "log(Phi(rho,z)) Analog N=100" or something like that?

Done!
image

> Demo02AnalogVsContinuousWithReflectance: on the top plot, the y-axis label should just be "log(R(rho)) [mm-2])" and then a length with red line = analog, blue line=CAW somewhere. On bottom plot, I would prefer "Analog" instead of "AAW" because "AW" stands for "absorption weighting" and analog has no absorption weighting.

Thanks, fixed! (though, isn't Analog just a binary "all at once" weighting?)
image

hayakawa16:
Regarding forward solver demos:  thanks for addressing all of my comments!  Just a few responses.
1) I did not realize that Demo04FluenceOfRhoAndZAndFt put out 3 plots.  Its hard to know what tabs are new when running in succession.  Now that I see the 3 plots, I see why the third is titled "Modulation".  Does everyone know that it is the ratio of the prior 2 plots?
2) glad my comments helped to debug :)
3) I'm going to see if I can debug the line issue with Demo07PhdOfRhoAndZTwoLayer and Demo06FluenceOfRhoAndZTwoLayer
4) Steve Jacques chapter 5 in the purple book says: Total energy absorbed [J/cm3] can be computed when the energy [J] of the light source is given.  When talking about Monte Carlo solutions, he says: It is convenient to consider the Monte Carlo simulation as yielding the fractional density matrix of incident light absorbed, A [1/cm3], in response to one unit of delivered power or
energy. I think all of our models assume a normalized source.
5) Demo10ROfRhoAndTimeMulti looks good.

Regarding ShortCourse  folder:  thanks for all of the fixes, they look great!
Analog is indeed a binary estimator and so no weight modification is performed to the photon due to absorption. That is why I didn't like the name "AW" (absorption weighting) used with this random walk process.  DAW and CAW random walk processes disallow termination of the photon due to absorption and instead deweight the photon along its trajectory and so "AW" in their name makes sense.

Thanks for your quick turn around!

dcuccia:
Thanks. 

> Does everyone know that it is the ratio of the prior 2 plots?

Anyone reading the script will know. In the future I'd like to combine back in one, but suggest we first figure out how to resolve the colorbars issue.

> fractional density matrix of incident light absorbed, A [1/cm3], in response to one unit of delivered power or
energy.

So it's normalized for use with either power and energy, as we hypothesized, thanks!

Re: analog, I guess I was just saying it is binary absorption weighing: 0 if not terminated, 1 if so. But all fine w/ me re: naming!

dcuccia:
@hayakawa16 thanks for looking into this. I appreciate the thought, but I'm not sure I buy it. This integer computation gets much closer to the double-precision limit of accuracy, but also produces the same line:

```csharp
        var rhos = Enumerable.Range(1, 100).Select(i => ((double)(i * 2 - 1)) /10).ToArray(); // range of s-d separations in mm
        var zs = Enumerable.Range(1, 100).Select(i => ((double)(i * 2 - 1)) / 10).ToArray(); // range of depths in mm
```
(as does this, but it has the same double precision issue):
```csharp
        var rhos = MathNet.Numerics.Generate.LinearSpaced(100, 0.1, 19.9); // range of s-d separations in mm
        var zs = MathNet.Numerics.Generate.LinearSpaced(100, 0.1, 19.9); // range of depths in mm
```

And if you zoom in, the data are "real" (and changing along rho!), but an order of magnitude larger than the rest: 
image

hayakawa16:
I have another idea.  This is an SDA solution.  One of the key assumptions about when you can apply an SDA solution is if mus'>>mua.  The top layer OPs were set to mua=1, mus'=1.  This violates this assumption.  If you set the top layer OPs to mua=0.1, mus'=1, and bottom mua=0.01, mus'=1, then line is not there I think.  What do you think about that?

dcuccia:
Did a little more experimentation - it seems to only show up for me when the top layer optical properties are mua: 1, musp: 1. Change them slightly, and it will go away. Changing bottom layer properties doesn't matter, it seems.

hayakawa16:
I reverted my rhos and zs modifications, and modified OPs so that line is not there.  I pushed this fix.

dcuccia:
Ok. Is it worth documenting the bug as a separate ticket? Some singularity with the `TwoLayerSDAForwardSolver.StationaryFluence` function?

dcuccia:
And are we otherwise ready to approve/merge?

dcuccia:
Missed the comment you made earlier, sorry. Yes I agree, if we see it pop up again for some other SDA-congruent optical properties, then maybe it's worth tracking as a bug.

hayakawa16:
On the first comment, at first I thought we violated the assumption that the top layer needs to be greater than l* because the embedded source at l* deep needs to be in the top layer.  This is well documented in the header of this class and there is code to check for that assumption.  I'm not sure we can know a definite lower bound on mus'/mua when the solution becomes invalid, however I guess adding a check if mus'==mua (and I think this check should be done for both layers) could be done and adding documentation to the header in this regard.  I can create an issue for that if that is what you had in mind.

hayakawa16:
This check should be added to all SDA forward solvers if they don't already have it.

hayakawa16:
Issue generated and assigned to me.  I noticed that the check I made to ensure top layer thickness is thick enough is in each TwoLayerSDAForwardSolver method.  Might be a better way to do this going forward.  I might ask advice from those with stronger coding skills than mine as to best way to implement this check.

dcuccia:
Thanks Carole! FWIW, I don't think it's related to musp==mua or limits. The original settings of (1,1) were the only thing I could get to generate the problem, and again only for the top layer.

hayakawa16:
I think the check should be in the SDA forward solvers.  Just because we don't see a visual aberration, doesn't mean the solution is valid.

lmalenfant:
> And are we otherwise ready to approve/merge?

Once you and Carole are good with the changes, I will take one last pass and we should be good to merge!

dcuccia:
I think any other changes would be in a different branch. I _think_ we're good here. Carole?

hayakawa16:
My technical review is complete.  I'll let Lisa approve after her final pass.

lmalenfant:
I am unable to build the project, I think due to the static abstract in the  IDemoScript interface.

src/Vts.Scripting/ScriptHelper.cs
```diff
@@ -0,0 +1,132 @@
...
+    /// <summary>
+    /// Method to create a standard scatter chart from the specified x and y values using Plotly.NET
+    /// </summary>
+    /// <param name="xValues">The values for the x axis.</param>
+    /// <param name="yValues">The values for the y axis.</param>
+    /// <param name="xLabel">The label for the x axis. Optional.</param>
+    /// <param name="yLabel">The label for the y axis. Optional.</param>
+    /// <param name="title">The title of the chart. Optional.</param>
+    /// <returns>A `GenericChart` instance representing the scatter chart.</returns>
+    public static Plotly.NET.GenericChart.GenericChart ScatterChart(double[] xValues, double[] yValues, string xLabel = "", string yLabel = "", string title = "")
```
lmalenfant, line 48: I see that this, LineChart and HeatMap are not being used, will they be used at a later time, can we add unit tests for these later?

src/Vts.Scripting/Program.cs
```diff
@@ -0,0 +1,29 @@
+﻿// This is a top-level program that lets you run any of the demos in the Vts.Scripts project.
+// It is a good place to start if you want to run a demo, or if you want to write your own code.
+using Vts.Scripting;
+
+// to run a single demo, find the one you want by browsing to them, and run it:
+Vts.Scripting.ShortCourse.Demo02AnalogVsContinuousWithReflectance.RunDemo();
```
lmalenfant, line 6: Why do you have the using and the fully qualified path here? Can we remove Vts.Scripting?

dcuccia: So that when people uncomment the code at the bottom, it will work.

src/Vts.Scripting/ScriptHelper.cs
dcuccia: They're used throughout the demos. I'm not sure how/why we'd unit test them - they're the things we're skipping in our unit tests. They'll just be helper functions at the top of the notebooks to help with plotting.

src/Vts.Scripting/Program.cs
lmalenfant: I meant can we replace:

`Vts.Scripting.ShortCourse.Demo02AnalogVsContinuousWithReflectance.RunDemo();`

with:

`ShortCourse.Demo02AnalogVsContinuousWithReflectance.RunDemo();`

src/Vts.Scripting/ScriptHelper.cs
lmalenfant: They are greyed out in my solution, possibly because it doesn't build.

src/Vts.Scripting/Program.cs
dcuccia: Ah, left that in to match the others - MonteCarlo and ForwardSolvers have collisions without the fully qualified name.

lmalenfant:
@dcuccia can you look into this and then we should be ready to merge:

>I am unable to build the project, I think due to the static abstract in the IDemoScript interface.
>
>This is the error:
>`Using both 'static' and 'abstract' modifiers requires opting into preview features.`

dcuccia:
Not sure what's going on here, as the preview features are enabled, per the .csproj file. I thought it might just be your machine, but right-clicking on the project (and running "dotnet build") and have the same problem. But, if I click the green "Play" button with the Vts.Scripting project as the active project, it works - and, I can see the Debug assets are built with a "just now" timestamp. Can you let me know if the same is true for you?

dcuccia:
That said, I pushed a change that removes the contract for now, until we figure out what's going on. It's not essential to the use of this code, going forward, just a good guardrail for making new examples uniform (which will happen by copy-paste even without this).

lmalenfant (approved):
Phew! We made it!

>Not sure what's going on here, as the preview features are enabled, per the .csproj file. I thought it might just be your machine, but right-clicking on the project (and running "dotnet build") and have the same problem. But, if I click the green "Play" button with the Vts.Scripting project as the active project, it works - and, I can see the Debug assets are built with a "just now" timestamp. Can you let me know if the same is true for you?

Yes, that was the same for me.

PR #98, opened 2023-09-12
Fix the TearDown error

lmalenfant:
Added a "using" when accessing the DataPoints so it hopefully disposes correctly and fixes the TearDown error


lmalenfant:
@hayakawa16 I looked back at the history of this file and it looks like we have added and removed the Garbage Collector code several times. The last change was a commit by me to remove it (I think). I removed it again but this didn't fix it so I added a using when enumerating through the DataPoints because it was a suggestion and then this should clean up once it finished accessing. The error has gone for me now but I should probably do a clean clone to make sure.

hayakawa16 (approved):
I like the way you implemented it compared to the other options you describe in the "Notes on implementation...".  Nice solution!  I pulled a clean clone and it works great!

PR #99, opened 2023-09-14
Create build.yaml

lmalenfant:


lmalenfant:
I was able to get this pipeline working on the VTS repo (it took me a while!) This pipeline does build, test and code coverage. The code coverage and test are in the same command but I think they still work as expected. This means now that every time we create a PR, it will run the checks.

hayakawa16 (approved):
Nice success!

dcuccia:
Would like to review but it will take some time (sorry!)

lmalenfant:
@dcuccia I added you more for visibility than to review, I know you're busy right now.

Feel free to take a look later and if you have suggestions we can make modifications then.

PR #101, opened 2023-09-15
Update readme.md

lmalenfant:
I added a .github repository in the Virtual Photonics organization so we could put the VP information on the home page. I removed that information here and added the sonar cloud analytics badge to the readme.


hayakawa16 (approved):
Sounds good.

PR #104, opened 2023-09-30
Feature/102 fix the sonarcloud issues in iforwardsover

lmalenfant:
@hayakawa16 before I add reviewers to the PR, I would like to review the XML comments with you first, I think we should remove the comment that states "Overload of x function" because the fact that there are multiple method definitions, means that they are overloads so this comment seems redundant, what are your thoughts?


src/Vts.Scripting/MonteCarlo/Demo07pMCInversion.cs
hayakawa16, line 1: This update does not remove build warning for Demo07pMCInversion: The event Demo07PmcInversion.PmcForwardSolver.PropertyChanged is never used.  Will that be done in another update?

hayakawa16 (approved):
HUGE effort on the reorganization of the IForwardSolver.  Thank you!

src/Vts.Scripting/MonteCarlo/Demo07pMCInversion.cs
lmalenfant: I just updated the part that was affected by the IForwardSolver changes, do you think we should fix this also?

lmalenfant: I fixed this and a few other code quality suggestions in the file.

lmalenfant:
@dcuccia @hayakawa16 I plan to merge this PR tomorrow as it has been open for 2 weeks.

dcuccia:
> @dcuccia @hayakawa16 I plan to merge this PR tomorrow as it has been open for 2 weeks.

I'll look at it tonight

dcuccia (commented):
@lmalenfant can you summarize what we did on this ticket? Was it just re-organizing the order of functions and fixing comments? (E.g. no changes to method signatures/API?) If we changed the API, a) how did we choose to augment/remove methods, and b) have we tested the "loosely types" Matlab calls??

lmalenfant:
> Was it just re-organizing the order of functions and fixing comments? (E.g. no changes to method signatures/API?)

Yes exactly, no other changes.

dcuccia:
> > Was it just re-organizing the order of functions and fixing comments? (E.g. no changes to method signatures/API?)
> 
> Yes exactly, no other changes.

Thanks. Please go ahead with the merge. I did see the removal of implicit "target typed"  constructor calls in the solver demos. In my opinion, those are unnecessary additions that add verbosity (and decrease conciseness), but that's just a preference based on my experience with functional style programming.

PR #106, opened 2023-09-30
Feature/105 fix the 5 sonarcloud bugs

lmalenfant:
@hayakawa16 These changes should fix the 5 remaining bugs in the VTS. If you could review the CylinderTissueRegionToolbox change, that would be great.

So we can leave the discrete code in the NurbsForwardSolver, I added a public property that defaults to false (existing functionality) so it could potentially be set to true to access the discrete code. I could make this value internal if we don't want to expose it publicly.

### DIFF
--- a/src/Vts/Modeling/ForwardSolvers/NurbsForwardSolver.cs
+++ b/src/Vts/Modeling/ForwardSolvers/NurbsForwardSolver.cs
@@ -2,7 +2,6 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
-using MathNet.Numerics;
 using Vts.Extensions;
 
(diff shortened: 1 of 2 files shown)


hayakawa16 (approved):
Since NurbsForwardSolver is included in this PR, I brought in the GUI and GUI tests and made sure all is okay.  I was adding in the local Vts dependency to the Vts.Gui.Wpf.Test and it showed a red x on the existing Microsoft.Windows.Compatibility (6.0.6) so I deleted this dependency.  Then I got into NuGet manager and see there were 2 updates to Microsoft.NET.Test and NUnit3TestAdapter so I updated those.  I brought up the GUI and Nurbs FS ran okay and so did the unit tests.

PR #108, opened 2023-10-01
Feature/107 fix bugs security issues and maintenance issues in mccl and mcpp

lmalenfant:


src/Vts.MonteCarlo.CommandLineApplication/Program.cs
hayakawa16, line 1: Why is suggested update to local variable CPUCount name cpuCount and not cPuCount?  I found an example of local variable in pMCDatabase.cs of pMCDatabase and suggested name is pMcDatabase.

src/Vts.MonteCarlo.PostProcessor/Program.cs
hayakawa16, line 1: Thanks!

src/Vts.MonteCarlo.CommandLineApplication/Program.cs
lmalenfant: CPU is an acronym like API and are considered words like Count and Database, putting an uppercase P in the middle is not right, I guess it made that suggestion incorrectly, it should be pmcDatabase in camel case and PmcDatabase in Pascal case.

hayakawa16: I see.  Thanks for your response.

hayakawa16 (approved):
Looks good.

PR #109, opened 2023-10-04
Fixed all the Sonar Cloud issues in the Test project

lmalenfant:
Fixed all the code smells in the Test project, there were about 31 of them in total.


src/Vts.Test/MonteCarlo/Factories/DetectorFactoryTests.cs
hayakawa16, line 1: I had a question about the change on line 94: when you modify an "out" parameter from type IDetector to var, does it not loosen the constraint on the parameter?

lmalenfant: I don't believe so because it will only suggest using var if there is no ambiguity. This out value is not used in the unit test though so we should probably discard its value with and underscore.

src/Vts.Test/MonteCarlo/Tissues/InfiniteCylinderTissueRegionTests.cs
hayakawa16, line 1: Why the change on line 61-62 from Double.PositiveInfinity to double.PositiveInfinity?  Aren't they equivalent?

src/Vts.Test/MonteCarlo/Factories/DetectorFactoryTests.cs
```diff
@@ -139,13 +139,13 @@ public void Demonstrate_user_defined_detector_usage()
             };
             var simInput = new SimulationInput
                 {
-                    DetectorInputs = new[] { detectorInput },
+                    DetectorInputs = new IDetectorInput[] { detectorInput },
```
lmalenfant, line 142: I changed this here but it is suggesting this change elsewhere in this file, so I think I need to change it in a few more places. This is the reason behind the change:
https://www.jetbrains.com/help/resharper/2023.2/CoVariantArrayConversion.html

hayakawa16 (approved):
Thanks for your help!  They look good.

src/Vts.Test/MonteCarlo/Factories/DetectorFactoryTests.cs
hayakawa16: I like fix.

PR #111, opened 2023-10-05
Feature/110 fix the bug in EllipsoidTissueRegion

lmalenfant:
Removed the unreachable code by changing case 0: (no intersections) to default:

### DIFF
--- a/src/Vts/MonteCarlo/Tissues/EllipsoidTissueRegion.cs
+++ b/src/Vts/MonteCarlo/Tissues/EllipsoidTissueRegion.cs
@@ -196,7 +196,7 @@ namespace Vts.MonteCarlo.Tissues
             double zto;
             switch (numberOfIntersections)
             {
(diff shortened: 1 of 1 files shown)


src/Vts/MonteCarlo/Tissues/EllipsoidTissueRegion.cs
hayakawa16, line 1: Change "visually" looks good and all unit tests pass.  I started up a simulation with infile_ellip_FluenceOfRhoAndZ.txt and 100,000 photons to stress test code and make sure somehow flow doesn't flow out bottom of case.

hayakawa16 (approved):
I performed my test run with N=100,000 photons and a specified a true ellipse (not a sphere) to make sure code gets exercised.  All looks good.  It took 20 hours on my new laptop though!  I think the laptop goes to sleep and pauses during the simulation?  This does not happen on linux.

PR #128, opened 2023-11-09
Added unit tests for BinaryArraySerializerFactory

lmalenfant:
@hayakawa16 these are my additions for the unit tests for the BinaryArraySerializerFactory, if they look good to you, I can merge them into your branch.

I reorganized the BinaryArraySerializerFactory so that the overloads were together and added XML comments to the public method. I usually would add XML comments to the internal methods also but this file will probably replaced by the refactored one.


hayakawa16 (approved):
Really like your unit test!  I will have to view and learn.  Thanks for your help!

lmalenfant:
I'm merging these changes into your branch. There is a code smell but that is just the cognitive complexity on the nested loops which will hopefully get resolved when we move to the other code.

PR #141, opened 2024-04-18
Added check if Matlab installed before running powershell commands that rely on its installation.

hayakawa16:
I added powershell commands in BuildTestRelease.ps1 and BuildTestReleaseMCCL.ps1 to check if Matlab is installed and if not to not run any commands that rely on its installation.


lmalenfant (approved):
Nice!

janakarana:
I ran the latest build on my Mac. This is to confirm that the bug at the end is now fixed.

hayakawa16:
Thanks @janakarana!

PR #142, opened 2024-04-23
feature/139-address-deprecation-warning-when-building-vts-under-net-v6

lmalenfant:
- Removed all the obsolete methods and properties
- Made AsEnumerable private so that it is only used by the GetEnumerator method

@dcuccia I'm not sure if this is what you had in mind for removing the obsolete method, I tried other ways of implementing GetEnumerator() but this appeared to be the most straight-forward. If you have another suggestion, let me know.

### DIFF
--- a/src/Vts/IO/BinaryArraySerializer.cs
+++ b/src/Vts/IO/BinaryArraySerializer.cs
@@ -9,11 +9,6 @@ namespace Vts.IO
     public class BinaryArraySerializer
     {
         /// <summary>
-        /// array to be written or read
-        /// </summary>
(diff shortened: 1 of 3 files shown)


hayakawa16 (approved):
This looks good to me.  I'd be interested to see how @dcuccia responds.

dcuccia (commented):
@lmalenfant Thanks for doing this. Yes, that's the right direction - perhaps just move the body of AsEnuerable to GetEnumerator directly. No need for a separate method, IMO.

dcuccia:
I see. Well, all uses of the IEnumerable interface "uses" it. But up to you, no biggie.

PR #144, opened 2024-04-29
Hotfix for the build error

lmalenfant:
The scripting project was still referencing "AsEnumerable" removed this in the ScriptHelper


hayakawa16 (approved):
Thanks for fixing this @lmalenfant!

PR #149, opened 2024-05-24
Feature 148 Updates to PowerShell Scripts

lmalenfant:
Updated the PowerShell scripts so they can all run independently, added a script for just the VTS libraries


lmalenfant:
Is this new behavior or did it happen with the previous version? When you say it crashes Chrome, is that the browser that crashes?

hayakawa16:
It happened on previous version so not caused by your updates.  It crashed my browser chrome.  It runs fine when I run it remotely on a computer on campus because nothing else is running on that computer.  I don't think it should hold up merge.  I just wanted to mention it in case you saw anything that might be causing it.

PR #160, opened 2024-08-07
Added warning if diffusion parameters are such that mua>=musp.

hayakawa16:
Description of fix provided in related Issue.

Please let me know your thoughts on this fix.  Thanks!


hayakawa16:
I need to add a unit test for this added code.

lmalenfant (approved):
Although it is not a best practice to use Console.WriteLine in a library, we are using it in many other locations so for now, adding one more will not make a difference. We might want to introduce a better way of logging these warnings in the future.

hayakawa16:
Thanks for your review @lmalenfant!

PR #162, opened 2024-08-13
13 rfx tallies assume a homogeneous or layered tissue geometry

hayakawa16:


hayakawa16:
I also did some code cleanup since I was editing these files.  So more changes shown than the ones required for this fix.

src/Vts/MonteCarlo/DataStructuresValidation/SimulationInputValidation.cs
```diff
@@ -166,24 +171,25 @@ private static ValidationResult ValidateCombinedInputParameters(SimulationInput
                         {
                             case true when
                                 ellipsoid.Center.X != 0.0 && ellipsoid.Center.Y != 0.0:
```
lmalenfant, line 173: I'm seeing an issue here with the floating point comparison, it suggests using a range:
https://www.jetbrains.com/help/resharper/CompareOfFloatsByEqualityOperator.html

src/Vts/MonteCarlo/DataStructuresValidation/SimulationInputValidation.cs
```diff
@@ -166,24 +171,25 @@ private static ValidationResult ValidateCombinedInputParameters(SimulationInput
...
-                                return new ValidationResult(
-                                    false,
-                                    "Ellipsoid must be centered at (x,y)=(0,0) for cylindrical tallies",
-                                    "Change ellipsoid center to (0,0) or specify non-cylindrical type tally");
+                                    Console.WriteLine("Warning: off center ellipsoid in tissue with cylindrical detector defined: user discretion advised");
+                                    return new ValidationResult(
+                                    true,
+                                    "Warning: off center ellipsoid in tissue with cylindrical detector defined",
+                                    "User discretion advised: change ellipsoid center to (0,0) or specify non-cylindrical type tally");
                             case true when ellipsoid.Dx != ellipsoid.Dy:
```
lmalenfant, line 179: Same comparison issue as above:
https://www.jetbrains.com/help/resharper/CompareOfFloatsByEqualityOperator.html

lmalenfant (commented):
Both files have an unused using.

lmalenfant:
@hayakawa16 I didn't request changes because I'm not sure that the 2 issues I mentioned are problematic. I can approve if it is working as expected.

src/Vts/MonteCarlo/DataStructuresValidation/SimulationInputValidation.cs
hayakawa16: @lmalenfant, thanks for your review.  I can work on those.

src/Vts/MonteCarlo/DataStructuresValidation/SimulationInputValidation.cs
hayakawa16: I understand jetbrains concern.  In this particular case, I truly am comparing with 0.0 though.  If it is slightly off, ellipsoid will be off center.  Since it is a user input (not determined from code), I would prefer to keep this as is.

src/Vts/MonteCarlo/DataStructuresValidation/SimulationInputValidation.cs
hayakawa16: Here I agree (although still user input so shouldn't different), so have modified code.

lmalenfant (approved):

lmalenfant (changes requested):
I approved and then realized that the quality gate failed, I'm not sure how we can resolve, or why it didn't fail on the previous build.

hayakawa16:
I can try to fix on this branch.  I'll take a closer look tomorrow.  Also, my VS continued to act funny.  The Test Explorer window wasn't working correctly until I Task Master closed VS and then opened and reloaded sln.  Do you think it is possibly related to my VS problems I had today?

lmalenfant (approved):

hayakawa16:
Thanks @lmalenfant!  Still working on it though!

PR #192, opened 2025-04-15
Feature/191 fix the build issue with the benchmark configuration

hayakawa16:
I have a first pass at working Benchmark code.  Please give it a review.


dcuccia:
Why is Benchmark.Net needed as a package in Vts.dll? Shouldn't it only be in Vts.Benchmark.dll?

lmalenfant:
> Why is Benchmark.Net needed as a package in Vts.dll? Shouldn't it only be in Vts.Benchmark.dll?

Benchmark.NET is added to the VTS conditionally for the Benchmark configuration so the package is not in the release version of the Vts.dll. We use some Benchmark attributes in our code but they are only used when running Benchmark. It is possible to Benchmark externally to the DLL and we might want to look into doing it that way, but for the initial way we set it up, we needed the attributes.

dcuccia:
Got it, thanks!

hayakawa16:
I think I'm ready for a review now.

lmalenfant (approved):
Looks good, I'm approving the changes.

hayakawa16:
@janakarana, would you like to review what is done on this branch?  If you're too busy, no problem.  Just wanted to give you a chance to see the method used to time MCCL.

janakarana (approved):
Thanks for asking me to check the changes. 
Looks good.

hayakawa16:
Thank you @janakarana!

PR #198, opened 2025-06-18
Added overload to GetMusp that takes in lambda0.  Also cleaned up cod…

hayakawa16:
…e and corrected links in comments.  Added unit test for this new method.


hayakawa16:
When you have time @janakarana, please review my new method from a technical point of view.  I would like @lmalenfant review from a coding point of view because it was suggested that I add the missing enum types into the switch statement in the default case but I didn't think I should do this because code used by GUI.

hayakawa16:
> > I would like @lmalenfant review from a coding point of view because it was suggested that I add the missing enum types into the switch statement in the default case but I didn't think I should do this because code used by GUI.
> 
> This is one of those circular suggestions so if you change it, it will then state that they are unused so I think it is better to leave it as is.

I worried that the suggestion was to put the missing enums under default which puts out an exception.  I am glad you agree to leave it as is.  Thank you!

janakarana (approved):
Looks. good. I approve it.

hayakawa16:
Thank you @janakarana!

PR #200, opened 2025-07-01
Feature/197 powerlawscatterer option to define lambda0

hayakawa16:
I made an attempt at solving this by adding Property to PowerLawScatterer and overload to constructor.  In Tissue I added overload to GetMusp with parameter lambda0 which checks if Scatterer is PowerLawScatterer and if so call the overload with parameter lambda0.  Also needed to add 2 overloads to GetOpticalProperties, one for single wavelength and one for array of wavelengths.

I have not added unit tests.  Will do this once the code looks okay.

### DIFF
--- a/src/Vts/Modeling/Spectroscopy/Tissue.cs
+++ b/src/Vts/Modeling/Spectroscopy/Tissue.cs
@@ -93,12 +93,7 @@ namespace Vts.SpectralMapping
         /// <returns>The absorption coefficient Mua</returns>
         public double GetMua(double wavelength)
         {
-            double mua = 0.0;
-            for (int i = 0; i < Absorbers.Count; i++)
-            {
-                mua += Absorbers[i].GetMua(wavelength);
-            }
-            return mua;
+            return Absorbers.Sum(t => t.GetMua(wavelength));
         }
 
         /// <summary>
@@ -109,6 +104,21 @@ namespace Vts.SpectralMapping
         public double GetMusp(double wavelength)
         {
             return Scatterer != null ? Scatterer.GetMusp(wavelength) : 0;
+        }
+
+        /// <summary>
+        /// Returns the reduced scattering coefficient for a given wavelength
+        /// </summary>
+        /// <param name="wavelength">Wavelength</param>
+        /// <param name="lambda0">Wavelength normalization factor</param>
+        /// <returns>The reduced scattering coefficient Mus'</returns>
+        public double GetMusp(double wavelength, double lambda0)
+        {
+            if (Scatterer != null)
+            {
+                return Scatterer is PowerLawScatterer scatterer ? scatterer.GetMusp(wavelength, lambda0) : Scatterer.GetMusp(wavelength);
+            }
+            return 0;
         }
 
         /// <summary>
@@ -139,7 +149,7 @@ namespace Vts.SpectralMapping
         public OpticalProperties GetOpticalProperties(double wavelength)
         {
             var mua = GetMua(wavelength);
-            var musp = GetMusp(wavelength);
+            var musp = GetMusp(wavelength, 1000);
             var g = GetG(wavelength);
             var n = N;
             return new OpticalProperties(mua, musp, g, n);
@@ -148,18 +158,50 @@ namespace Vts.SpectralMapping
         /// <summary>
         /// Returns the optical properties for a given wavelength
         /// </summary>
+        /// <param name="wavelength">Wavelength</param>
+        /// <param name="lambda0">Wavelength normalization factor</param>
+        /// <returns>The optical properties</returns>
+        public OpticalProperties GetOpticalProperties(double wavelength, double lambda0)
+        {
+            var mua = GetMua(wavelength);
+            var musp = GetMusp(wavelength, lambda0);
+            var g = GetG(wavelength);
+            var n = N;
+            return new OpticalProperties(mua, musp, g, n);
+        }
+
+        /// <summary>
+        /// Returns the optical properties for an array of wavelengths
+        /// </summary>
         /// <param name="wavelengths">Wavelength</param>
         /// <returns>The optical properties</returns>
         public OpticalProperties[] GetOpticalProperties(double[] wavelengths)
         {
             var opArray = new OpticalProperties[wavelengths.Length];
-            for (int i = 0; i < wavelengths.Length; i++)
+            for (var i = 0; i < wavelengths.Length; i++)
             {
                 opArray[i] = GetOpticalProperties(wavelengths[i]);
             }
             return opArray;
         }
+
+        /// <summary>
+        /// Returns the optical properties for an array of wavelengths
+        /// </summary>
+        /// <param name="wavelengths">Wavelength</param>
+        /// <param name="lambda0">Wavelength normalization factor</param>
+        /// <returns>The optical properties</returns>
+        public OpticalProperties[] GetOpticalProperties(double[] wavelengths, double lambda0)
+        {
+            var opArray = new OpticalProperties[wavelengths.Length];
+            for (var i = 0; i < wavelengths.Length; i++)
+            {
+                opArray[i] = GetOpticalProperties(wavelengths[i], lambda0);
+            }
+            return opArray;
+        }
     }
+
     /// <summary>
     /// tissue provider class
     /// </summary>
--- a/src/Vts/Modeling/Spectroscopy/PowerLawScatterer.cs
+++ b/src/Vts/Modeling/Spectroscopy/PowerLawScatterer.cs
@@ -4,9 +4,10 @@ namespace Vts.SpectralMapping
 {
     /// <summary>
     /// Returns scattering values based on Steve Jacques' Skin Optics Summary:
-    /// https://omlc.ogi.edu/news/jan98/skinoptics.html
+    /// https://omlc.org/news/jan98/skinoptics.html
     /// This returned reduced scattering follows the approximate formula:
-    /// mus' = A1*lamda(-b1) + A2*lambda(-b2)
+    /// mus' = A1*(lamda/lambda0)(-b1) + A2*(lambda/lambda0)(-b2)
+    /// with default value of lambda0=1000nm
     /// </summary>
     public class PowerLawScatterer : BindableObject, IScatterer
     {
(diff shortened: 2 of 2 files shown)


src/Vts/Modeling/Spectroscopy/Tissue.cs
```diff
@@ -148,18 +159,50 @@ public OpticalProperties GetOpticalProperties(double wavelength)
         /// <summary>
         /// Returns the optical properties for a given wavelength
         /// </summary>
+        /// <param name="wavelength">Wavelength</param>
+        /// <param name="lambda0">Wavelength normalization factor</param>
+        /// <returns>The optical properties</returns>
+        public OpticalProperties GetOpticalProperties(double wavelength, double lambda0)
```
lmalenfant, line 165: Since GetOpticalProperties is not in the interface, I think you could combine these 2 methods by passing the lambda0 as an optional parameter like we did for GetMusp initially. Then if lambda0 is NaN we call GetMusp without that parameter, otherwise we call it with the parameter.

src/Vts/Modeling/Spectroscopy/Tissue.cs
```diff
@@ -148,18 +159,50 @@ public OpticalProperties GetOpticalProperties(double wavelength)
...
             return opArray;
         }
+
+        /// <summary>
+        /// Returns the optical properties for an array of wavelengths
+        /// </summary>
+        /// <param name="wavelengths">Wavelength</param>
+        /// <param name="lambda0">Wavelength normalization factor</param>
+        /// <returns>The optical properties</returns>
+        public OpticalProperties[] GetOpticalProperties(double[] wavelengths, double lambda0)
```
lmalenfant, line 195: Same comment as above, since GetOpticalProperties is not in the interface, I think you could combine these 2 methods by passing the lambda0 as an optional parameter like we did for GetMusp initially. Then if lambda0 is NaN we call GetMusp without that parameter, otherwise we call it with the parameter.

lmalenfant (commented):
I didn't notice these methods before, hopefully I'm correct about being able to have the optional parameter.

src/Vts/Modeling/Spectroscopy/Tissue.cs
hayakawa16: I have pushed my updates.

src/Vts/Modeling/Spectroscopy/Tissue.cs
```diff
@@ -128,38 +139,53 @@ public double GetG(double wavelength)
...
         /// </summary>
         /// <param name="wavelengths">Wavelength</param>
+        /// <param name="lambda0">Optional wavelength normalization factor</param>
         /// <returns>The optical properties</returns>
-        public OpticalProperties[] GetOpticalProperties(double[] wavelengths)
+        public OpticalProperties[] GetOpticalProperties(double[] wavelengths, double lambda0 = double.NaN)
         {
             var opArray = new OpticalProperties[wavelengths.Length];
-            for (int i = 0; i < wavelengths.Length; i++)
+            if (double.IsNaN(lambda0))
```
lmalenfant, line 169: We could move this if statement inside the loop, it would be even less duplicated code.

I asked CoPilot if it thought that was a good solution but it said not, that it was good the way it was written, so I asked about duplication and it came up with this solution:
`for (var i = 0; i < wavelengths.Length; i++)
{
    opArray[i] = double.IsNaN(lambda0)
        ? GetOpticalProperties(wavelengths[i])
        : GetOpticalProperties(wavelengths[i], lambda0);
}`

lmalenfant: After giving CoPilot both options, it finally agreed with me that the code above is the best option:
image

hayakawa16: Hi @lmalenfant, I see that your proposed change is better for code duplication, but it is better for performance?  I considered performance when I put the "if" outside the "for loop".  Going through an "if" within a for loop when the outcome of the "if" is the same for each instance of the "for" seems like additional processing.  Does CoPilot consider performance?  I'd be interested what it says.

lmalenfant: It does, if you look at the Recommendation, it is taking that into account and it said the difference is insignificant. However, if you think this could affect performance in a negative way due to the way we utilize the code, I am good leaving it as is. We could do some performance checks of our own later when we have more time.

hayakawa16: Tissue is not referenced by any class in the Vts except the unit tests.  It is referenced by the GUI though.  I think the timing difference will show when numerous wavelengths on Spectral Panel specified.  I'll give this a try.

hayakawa16: I found no difference in performance with the GUI.  In addition, I found documentation that branch prediction will usually make sure that there is no difference at all in these cases, and the code is more readable.
https://stackoverflow.com/questions/38741599/using-an-if-statement-inside-loop-vs-outside-the-loop
So I have pushed your suggested code modification.  Thank you @lmalenfant !

hayakawa16:
Arg!  I inadvertently committed from Vts that was integrated into the Vts.Gui.Wpf.  So I had two sources committing to same branch.  I reverted commit on Wpf vts, cloned a clean clone of Vts, updated to branch, reapplied commit, and pushed.  It worked!

hayakawa16:
Hi @dcuccia, thank you for chiming in.  If I understand your comment correctly, you feel that there is no need for a lambda0 additional parameter because it can be taken into account by modifying the up-front A coefficient?  If so, I see your point.
I can think of two reasons to allow users to specify this lambda0:
1) The formulation of the Power Law scatterer is often defined with this wavelength normalization factor.   Our code in PowerLawScatterer hardcodes lambda0=1000nm and we cite OMLC because OMLC original web pages provided this value?  However this value is not listed on the current page.  This publication: https://omlc.org/news/dec14/Jacques_PMB2013/Jacques_PMB2013.pdf
uses 500nm as the normalization factor in Equations (1) and (2).  So this form:
mus'(lambda)=A(lambda/lambda0)^(-b)
of the equation is standard, but the value of lambda0 is not.  In addition, if a user wanted to use Jacques' equations and specify lambda0=500nm, they could not directly, they'd have to use your scaling.
2) This modification arose because both @janakarana and I have been solving the inverse problem which tries to determine the coefficients A and b or A1, b1, A2, b2 in the extended Power Law:
A1(lambda/lambda0)^(b1)+A2(lambda/lambda0)^b2.
What we have found is that if you select lambda0 to be around the midpoint of the wavelengths you are using in the inversion, the solution has better success.  Often since the measured data is not perfect (it does not obey the Power Law perfectly), altering lambda0 has influence over the recovered A and b and so playing around with various lambda0 has merit. Now we *could* determine the appropriate A1' and A2', however it is more straight-forward to use another lambda0 value and solve for A and b directly because initial guesses and final converged values make more intuitive sense since they are consistent with the standard form of the Power Law scatterer equation.

lmalenfant (approved):

hayakawa16:
Thanks @lmalenfant for your approval!  @dcuccia, any responses to my comments above?

lmalenfant:
@dcuccia we didn't change the interface, we changed one of the constructors for PowerLaw only.

dcuccia:
@lmalenfant that's great! And, is there a uniform implementation for all uses of PowerLawScatterer functions, with the class-wide normalization parameters used in the calculations (default center wavelength, or custom)?

hayakawa16:
Hi @dcuccia, in PowerLawScatterer class we have added overloads to the constructor and GetMusp to enable default center wavelength or custom. Within the Vts the only class calling PowerLawScatterer is the Tissue class (within the SpectralMapping namespace).  We have added overloads to the methods in Tissue class for instance, GetOpticalProperties and GetMusp, that enable default center wavelength or custom.
Is that what you were interested in?

hayakawa16:
This is example python code I wrote that drove some of the changes:
    # Define parameters to fit [Hb,HbO2,A,b]
    measuredData = [28.4, 22.4, 1.2, 1.42]
    # Construct absorber
    chromophoresMeasuredData = Array.CreateInstance(IChromophoreAbsorber, 2)
    chromophoresMeasuredData[0] = ChromophoreAbsorber(ChromophoreType.HbO2, measuredData[0])
    chromophoresMeasuredData[1] = ChromophoreAbsorber(ChromophoreType.Hb, measuredData[1])
    # Construct scatterer
    scattererMeasuredData = PowerLawScatterer(measuredData[2], measuredData[3])
    # len(opsMeasured)=4 per number of wavelengths
    lambda0=750.0
    opsMeasured = Tissue(chromophoresMeasuredData, scattererMeasuredData, "", n=1.4).GetOpticalProperties(wavelengths, lambda0)

I needed to be able to specify lambda0 and also obtain the optical properties via Tissue.  GetOpticalProperties calls GetMusp.

hayakawa16:
I feel your suggestion.  I created fix from bottom up.  I think you are suggesting top down addition of lambda0. By taking lambda0 in constructor, did you mean PowerLawScatterer constructor **and** Tissue constructor?  I thought taking lambda0 in Tissue constructor didn't make sense because IScatterer could be other than PowerLawScatterer.  This is where I get stuck.

dcuccia:
From the C# API perspective:
```csharp
IChromophoreAbsorber[] abs = [
    new ChromophoreAbsorber(ChromophoreType.HbO2, measuredData[0]),
    new ChromophoreAbsorber(ChromophoreType.Hb, measuredData[1]),
];
var (A, b, lambda0) = (measuredData[2], measuredData[3], 750f);
var scat = new PowerLawScatterer(A, b, lambda0); // lambda0 default for constructor can be whatever you want (sometimes best at 1nm for historical compatibility...)
var tiss = new Tissue(abs, scat, "", 1.4);
tiss.GetOpticalProperties(wavelengths); // calls scat.GetOpticalProperties(wavelengths) under the hood, "doesn't care" how scat does it's job...
```

hayakawa16:
@dcuccia, thanks for your help!  I understand concept.  I'll try again.

hayakawa16:
@dcuccia, this is a much improved solution.  Now the dependency of PowerLawScatterer on lambda0 is injected upon instantiation of this class as it should be.  Overloads in Tissue removed.  Thanks for your help!

lmalenfant (approved):

hayakawa16:
Thank you @lmalenfant!

dcuccia:
Sounds great

PR #203, opened 2025-08-25
Bug/202 unexpected focusing effect in infinite cylinder in vts

hayakawa16:
Don't require a review yet.  Just want to view changes with respect to main branch.


hayakawa16:
I'm ready for a review now.

lmalenfant (approved):
Looks good!

